PR #1, opened 2021-04-29
add mznt on to the device

sherimickelson:


src/turb.F
```diff
@@ -1107,10 +1107,11 @@ subroutine sfc_and_turb(getsfc,getpbl,nstep,dt,dosfcflx,cloudvar,qbudget,    &
     call comm_2d_corner(s1)
     call comm_2d_corner(znt)
   if( sfcmodel.eq.4 )then
+    !$acc update device(mznt)
```
johnmauff, line 1110: Should the update device(mznt) directive occur before or after the comm_2d_corner() call

PR #4, opened 2021-06-02
Changes to maxmin.F to remove pcast errors.

sherimickelson:


src/maxmin.F
```diff
@@ -249,7 +257,6 @@ subroutine maxmin2d(izz,jzz,f,nstat,rstat,amax,amin)
 
     if(myid.eq.0)then
 #endif
-      !$acc end data
       write(6,100) amax,fmax,imax,jmax,1,    &
                    amin,fmin,imin,jmin,1
 100   format(2x,a6,':',1x,g13.6,i5,i5,i5,    &
```
johnmauff, line 262: I think that the values in this write statement which is executed on the CPU will not be updated.

sherimickelson: Thanks @johnmauff .  I'll revert these changes and test the code again.  I'll create a new PR if everything looks good.

PR #19, opened 2021-08-19
Port of fallout and satadj to GPU

sherimickelson:


johnmauff (commented):
What kind of speedup are you achieving with your GPU port for each subroutine

sherimickelson:
This was what I saw for fallout:
Before port:
```fallout  :        5.26    1.12%```
After port:
``` fallout  :        2.47    0.54%```
I ended up on a busy node for the satadj port so my numbers might not reflect accurate performance.

johnmauff:
Bummer that is not the speedup I was  hoping for.  At least that is an improvement over the CPU resident version.  What about satadj?

sherimickelson:
I ended up on a busy node so I'm not sure I trust these timings, but here's what I saw for satadj:
Before port:
```satadj   : 3.91    0.97%```
After port:
```satadj   : 1.50    0.33%```

PR #20, opened 2021-08-24
Optimization work for the fallout subroutine

sherimickelson:
Optimization of the fallout_gpu subroutine
Original timing:
```fallout  :        5.87    9.34%```
New timing:
```fallout  :        1.00    1.77%```
Timings use default grid and faster options.

### DIFF
--- a/src/kessler.F
+++ b/src/kessler.F
@@ -737,16 +737,17 @@
       real, dimension(0:nk) :: ffk
       real, dimension(nk) :: qtmp
 
-      integer :: nfall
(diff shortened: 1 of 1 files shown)


johnmauff (commented):
I see that you are now parallelizing over the i,n, and k loop.  Why are you not parallelizing over the j loop?

sherimickelson:
I wasn't able to get it working just right and it'll take a little bit more work.  My plan was to split this optimization in 2 parts in order to get this first set  into the code base sooner.

PR #24, opened 2021-12-01
New location search implementation for droplet parcels

sjsprecious:
For a particular parcel, the current location search implementation in the `droplet_driver` subroutine always started from the end of a dimension and scanned backwards until the parcel was found. Such a linear search strategy at every time step was computationally costly and not necessary since a parcel would not move too far.

According to @johnmauff's suggestions, it is better to store the previous location of a parcel and use that as the starting search point for the next time step. In addition, the search range could be restricted to a small value with the assumption that a parcel does not move too far from one time step to another.

This PR introduces the new implementation following @johnmauff's suggestions. We also add a verification test and could confirm that the new implementation returns the identical location indexes as the original implementation.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -74,16 +74,16 @@ FC   = mpif90
 CPP  = cpp -C -P -traditional -Wno-invalid-pp-token -ffreestanding
 #
 # -------PERF-------
-#OPTS = -Mfree -O2 -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel -ta=tesla:cc70,lineinfo,nofma
-#OPTS = -Mfree -O2 -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Minfo=accel -ta=tesla:cc70,lineinfo
-OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,math_uniform -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Minfo=accel
+#OPTS = -Mfree -O2 -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel -ta=tesla:cc70,lineinfo,nofma -Kieee
+#OPTS = -Mfree -O2 -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Minfo=accel -ta=tesla:cc70,lineinfo -Kieee
+OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,math_uniform -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Minfo=accel -Kieee
 #DM   = -DMPI -D_OPENACC 
 # -------END PERF------
 #
 # -------DEBUG------
-#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mnoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel
-#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel
-DM   = -DMPI -D_OPENACC -D_B4B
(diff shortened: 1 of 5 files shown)


src/Makefile
```diff
@@ -74,16 +74,16 @@ FC   = mpif90
...
 #DM   = -DMPI -D_OPENACC 
 # -------END PERF------
 #
 # -------DEBUG------
-#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mnoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel
-#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel
-DM   = -DMPI -D_OPENACC -D_B4B
+#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mnoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel -Kieee
+#OPTS = -Mfree -g -O2 -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform -Ktrap=none -Mautoinline -Minline=reshape -Mpcast -Mnofma -Minfo=accel -Kieee
+DM   = -DMPI -D_OPENACC -D_VERIFY_FIND_LOC # -D_B4B
```
sherimickelson:
```suggestion
#DM = -DMPI -D_OPENACC -D_B4B
DM = -DMPI -D_OPENACC -D_VERIFY_FIND_LOC
```

sherimickelson (commented):
I think these modifications look good and I only have one minor suggestion.  Thanks for these contributions!

src/parcel.F
```diff
@@ -1076,14 +1086,55 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
 
         !print *,'droplet_driver: point #4'
         kflag = 1
+! JS - jflag could be negative somehow, which will break
+!        the find_vertical_location_index subroutine;
+!        use the original interface but with the 
+!        pdata_locind(:,3) here
+!    - z3d could be NaN somehow, reset kflag = 1 to be 
+!        consistent with the original implementation
+#ifdef _VERIFY_FIND_LOC
```
johnmauff, line 1094: Long term I don't think we want to maintain the _VERIFY_FINE_LOC ifdef in the source code.  Maybe we perform the commit so that it is in the gpu-opt branch and then subsequently delete it during code cleanup.  

Alternatively, could we find a way to maintain the verification code without adding significant complexity to the source code.

src/cm1.F
```diff
@@ -2207,6 +2209,8 @@ program cm1
       pdata = 0.0
       allocate(   ploc(nparcels,  3   ) )
       ploc = 0.0
+      allocate(  pdata_locind(nparcels,3) )
+      pdata_locind = -100
```
johnmauff: There are multiple locations for the setting of -100 to be an undefined value.  Could we do this in a single location so that it does not accidentally get changed to something else and then it breaks the code?

src/parcel.F
sjsprecious: Thanks for this comment. I agree that it is better to keep this verification test in the `gpu-opt` branch for future reference but remove it when we issue the PR to the `main` branch.

I think another alternative way could be generating a `verification` branch which keeps all the current and future verification tests, and cleaning up everything in the `gpu-opt` branch. We could consistently sync the `verification` branch whenever there is an update in the `gpu-opt` branch.

src/cm1.F
sjsprecious: Do you mean that some variables/arrays are uninitialized in CM1? Maybe we could write an `initialization` subroutine and always call this subroutine when we want to initialize a scalar/1-D/2-D/3-D array? In this way we could set `-100` in that subroutine once for the whole program.

johnmauff: I would suggest that you create a parameter with a name like "undefined_index" in a single location in the source code and then you test against it in the parcel.F90 file.

sjsprecious: Thanks @johnmauff for your detailed suggestion. I have added a parameter named `undefined_index` in the `constants.F` source code and defined it as `-100`. Other codes are updated correspondingly.

johnmauff:
Thanks for addressing the feedback.

PR #25, opened 2022-01-14
Modifications for gpu testing framework

sherimickelson:
Modified the Makefile to be more flexible
 - options include: FC, USE_MPI, USE_OPENMP, USE_DOUBLE, USE_OPENACC, DEBUG
 - example: ``FC=pgf90 USE_MPI=false USE_OPENMP=false USE_DOUBLE=false USE_OPENACC=false DEBUG=true``

Added the ability to read in a specified namelist file.  This is done through an optional command line argument ``--namelist`` The default is "namelist.input".  Example of how to read in a different namelist: ``--namelist namelist.asd04-128x1k.input``

PR #26, opened 2022-02-01
Fix multi-MPI-GPU runtime error

sjsprecious:
The CM1 code on the `gpu-opt` branch would fail the run with multiple MPI ranks and GPUs per node. The reason is that at line https://github.com/george-bryan/CM1/blob/gpu-opt/src/parcel.F#L1100, the access to **zf(iflag,jflag,kflag)** is invalid when **jflag=-100**. However, the old CM1 code with the less efficient location search scheme only needs to access **zf(iflag,jflag,kflag+1)** (like https://github.com/george-bryan/CM1/blob/gpu-opt/src/parcel.F#L1103) and somehow this access is valid even if **jflag=-100** in the old code as well. Therefore, we just revert the location search scheme at this particular section to the old implementation so that we could move on with other PR (https://github.com/george-bryan/CM1/pull/25).

A more fundamental question here is why **jflag** would be **-100** here and if this is not a bug from the original implementation, maybe we need to have a better way to handle this condition.

### DIFF
--- a/src/parcel.F
+++ b/src/parcel.F
@@ -1089,6 +1089,7 @@
 !        the find_vertical_location_index subroutine and 
 !        lead to invalid access to zf(iflag,jflag,kflag);
 !        revert to the old implementation
+        if( .not. terrain_flag )then
           do while( z3d.gt.zf(iflag,jflag,kflag+1) )
             kflag = kflag+1
           enddo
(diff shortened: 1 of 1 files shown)


johnmauff (approved):
I think this looks like a reasonable response to what appears to be a code bug that is unrelated to your new index search scheme.

sherimickelson:
Thanks for looking into this and this seems reasonable

sherimickelson (approved):

PR #28, opened 2022-02-03
David physics

drichte2:
This is a restructuring of the parcels. To keep all of the droplet stuff together, I have made "droplet.F", which contains the original droplet_driver. In addition to this, I have now created a new subroutine called "rk2_integration" which is the old rkloop simply removed and put here. This will allow me to continue developing the implicit solver "BE_integration" once this gets merged


johnmauff (approved):
Thanks for the pull request. It looks good to me.

PR #29, opened 2022-02-08
Rewrite the Makefile 

sjsprecious:
The Makefile is rewritten so that it could be run either by the `make` command or the CM1 verification framework (https://github.com/sherimickelson/CM1_benchmarkingWF). The detailed options remain the same for a particular configuration but the new structure is easier for us to include further changes.

Currently only the `nvhpc` and `pgi` compilers are tested on Casper, which compile the CM1 codes successfully.


src/Makefile
```diff
@@ -1,271 +1,155 @@
...
-ifeq (${USE_MPI_L},false)
-ifeq (${USE_OPENMP_L},false)
-#  single processor, Intel compiler 
-#      (eg, NCAR's yellowstone/caldera/cheyenne)
-override FC   = ifort
-OPTS = -g -O3 -xHost -ip -assume byterecl -fp-model precise -ftz -no-fma -init=snan,arrays -traceback
-CPP  = cpp -C -P -traditional -Wno-invalid-pp-token -ffreestanding
+#  NVHPC compiler
+#-----------------------------------------------------------------------------
+ifeq (${FC},nvfortran)
```
johnmauff, line 47: @sjsprecious  While this makefile is more readable versus the previous version I think there is still room for improvement.  In particular, I think that indents could be used so as to make the if structure much easier to understand.

sjsprecious: @johnmauff Thanks for the comments. Just add the indents.

johnmauff: Thanks, this looks good

sherimickelson (approved):
@sjsprecious , thanks.  I think this looks good and thanks for the improvements.

PR #31, opened 2022-02-09
Attempt 2 at putting into droplet.F

drichte2:
Okay, in a new branch I have now re-done the droplet.F separation, and broke "rk2_integration" into its own subroutine, starting from the most up-to-date gpu-opt version. I have not yet looked at the jflag = -100, that's on the agenda tomorrow


johnmauff (approved):
Looks good from a visual inspection.  Will run the verification test in the next couple of days to verify that answers agree.

PR #32, opened 2022-02-09
Convert 'routine vector' to 'routine seq'

johnmauff:
This pull request merges in some changes indicates that functions are sequential versus vector.  This change will decrease execution time by approximately 8%.  This also adds a verification namelist for the ASD03 configuration


src/Makefile
```diff
@@ -29,10 +29,10 @@ LINKOPTS      =
...
 #  Note:  you may need to remove -DNCFPLUS if you use an old version of netcdf
 #-----------------------------------------------------------------------------
-OUTPUTINC     += -I$(NETCDFBASE)/include
-OUTPUTLIB     += -L$(NETCDFBASE)/lib
-OUTPUTOPT     += -DNETCDF -DNCFPLUS
-LINKOPTS      += -lnetcdf -lnetcdff
+#OUTPUTINC     += -I$(NETCDFBASE)/include
+#OUTPUTLIB     += -L$(NETCDFBASE)/lib
+#OUTPUTOPT     += -DNETCDF -DNCFPLUS
+#LINKOPTS      += -lnetcdf -lnetcdff
```
sjsprecious: So we want to turn off these options by default? Maybe we could add a `USE_NETCDF_L` option?

src/comm.F
```diff
@@ -257,7 +257,7 @@ subroutine getcorner_GPU(s,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2)
       integer reqs(8)
       integer tag1,tag2,tag3,tag4
       integer, dimension(mpi_status_size,8) :: status1
-      logical, parameter :: Debug =.TRUE.
+      logical, parameter :: Debug =.FALSE.
```
sjsprecious, line 260: Can we move `Debug` parameter to a shared module so that we only need to change its value once?

src/domaindiag.F
```diff
@@ -9692,7 +9692,7 @@ subroutine domaindiag(nstep,mtime,nwritet,trecs,trecw,qname,qunit,dt,dosfcflx, &
         vardesc(nvar) = 'maximum (grnd-rel) horiz wind speed'
         varunit(nvar) = 'm/s'
         vargrid(nvar) = 's'
-        print *,'domaindiag: wspmax'
+        !print *,'domaindiag: wspmax'
```
sjsprecious, line 9695: Do we need to keep these `print` statements?

src/parcel.F
```diff
@@ -2447,55 +2448,55 @@ subroutine write_prclctl(name_prcl,desc_prcl,unit_prcl,prec)
       end subroutine write_prclctl
 
       real function calcW1(rx,ry,rz)
-      !$acc routine vector 
+      !$acc routine vector
```
sjsprecious, line 2451: why use `vector` instead of `seq` here?

sjsprecious (changes requested):

src/Makefile
johnmauff: Yes, I think it would be a good idea to add a USE_NETCDF_L flag.

src/comm.F
johnmauff: I added the Debug flag to turn on print statements when I was looking at issues with the GPU-direct.  I don't envision that this variable will be around long term, just until the code is fully debugged.

src/domaindiag.F
johnmauff: Again, I am still debugging the domaindiag.F routine which is why there is still a number of random printstatements.

src/parcel.F
johnmauff: I wanted to leave those 'routine vector' in the code because I did not want to cause conflicts with David's merge.  I can easy clean these up now that David's changed have gone in.

src/Makefile
sjsprecious: And at the same time, maybe add some instructions somewhere about how to use the new Makefile?

src/comm.F
sjsprecious: I see. Then we could clean it up when the code is in a good shape.

src/domaindiag.F
sjsprecious: Gotcha!

src/Makefile
johnmauff: Yes, I think both of those would be a good idea.  Do you want to take those on?

src/parcel.F
sjsprecious: That makes sense. Hope we remember to revisit this code after David's PR.

src/Makefile
sjsprecious: Sure. Let me see if I could directly make changes on your branch. Otherwise, I would just make a code suggestion and you could add it later.

sjsprecious:
@johnmauff I have updated the `README-GPU.md` and `Makefile` files to include instructions about how to use the new Makefile. Hope it looks better now and let me know if you think more changes are needed for other files.

sjsprecious (approved):

PR #35, opened 2022-02-23
David physics 2

drichte2:
I have made a few more somewhat large changes, adding an implicit droplet solver which is now working. It shouldn't change much of what was already there, but only adds all the necessary stuff for subroutine "BE_integration"


src/Makefile
```diff
@@ -138,7 +138,8 @@ endif
 #  Intel compiler
 #-----------------------------------------------------------------------------
 ifeq (${FC},ifort)
-    OPTS         += -g -O3 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback
+#    OPTS         += -g -O3 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback
```
sjsprecious: I also realized the `-init=snan,arrays` flag would broke the CM1 run on NCAR's machine using the Intel compiler. I would suggest to removing this line and keeping the updated version only.

src/droplet.F
```diff
@@ -394,13 +401,13 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
           iflag = 1
         ELSE
           ! cm1r19:
-          call find_horizontal_location_index (pdata_locind(np,1), x3d, ib, ie+1, xf, ni+2, iflag)
+          call find_horizontal_location_index (pdata_locind(np,1), pdata(np,prx), ib, ie+1, xf, ni+2, iflag)
```
sjsprecious: Could @drichte2 explain why using `pdata(np,prx)`, `pdata(np,pry)` and `pdata(np,prz)` instead of using `x3d`, `y3d` and `z3d` here?

src/solve2.F
```diff
@@ -2272,7 +2272,7 @@ subroutine solve2(nstep,rbufsz,num_soil_layers,                  &
 #ifdef MPI
       ! finish comms for cflquick:
       call mpi_wait(reqc,mpi_status_ignore,ierr)
-      print *,'solve2: cflmax is: ',cflmax
+      !print *,'solve2: cflmax is: ',cflmax
```
sjsprecious: could we just delete this line instead of commenting it out?

src/droplet.F
drichte2: This is related to the jflag issue -- at this point, these are identical so it technically doesn't matter which one you use. I switched to "pdata" because these contain all the new values. i.e., rk2_integration and BE_integration end with applying the parcel boundary conditions and then setting pdata(np,prx)=x3d, pdata(np,pry)=y3d, pdata(np,prz)=z3d.

src/solve2.F
drichte2: Sounds good to me. Someone else had put it in there to debug, and it was printing out way too much stuff

src/droplet.F
sjsprecious: Thanks @drichte2 . That makes sense to me.

src/droplet.F
```diff
@@ -1234,268 +1123,563 @@ subroutine rk2_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,r
...
+         num100 = num100+1
+
+         call LV_solver(diffnorm,dt_nondim,taup_scale,rt_start,rt_zeros,flag,prsval,rhoval,tval,qval,pdata(np,prtp),pdata(np,prms),pdata(np,prrp))
+
+      end if
+
+      if (flag==1) num1000 = num1000+1
+
+      !In rare cases the nonlinear solvers fail to converge
+      if (isnan(rt_zeros(1)) .OR.  isnan(rt_zeros(2)) .OR. rt_zeros(2)<0) then
```
sjsprecious: According to this thread (https://forums.developer.nvidia.com/t/error-isnan-has-not-been-explicitly-declared/132718/6), `isnan` is not a standard extension and may not be supported by some compilers like `PGI`. I would suggest using a better way to detect the NaN value here.

sherimickelson (approved):

src/droplet.F
johnmauff: Has the isnan() issue been resolved yet?

sjsprecious: Not yet. I plan to work on this after we have a stable verification framework and namelist.

johnmauff (approved):

src/Makefile
sjsprecious: This compiler flag should be kept and the runtime error is fixed by George in the latest `gpu-opt` branch.

src/droplet.F
sjsprecious: Updated by @drichte2

sherimickelson (approved):

johnmauff (approved):

sjsprecious (approved):

PR #37, opened 2022-03-10
Fix the metric output from the CM1 GPU code

sjsprecious:
Using the namelist `namelist.asd03-verify.input`, the CM1 CPU code returns zero LWP, precipitation and cloud fraction metric output. However, the output is non-zero when using the CM1 GPU code.

It turns out that this issue is caused by putting the scalar variables in the `acc data create` directive. After removing those scalar variables, the GPU code returns zero LWP, precipitation and cloud fraction, consistent with the CPU results.

All the revisions in this PR is based on the version from the `gpu-opt` branch (commit: bd7f069638a626b18266ad5b7b12f4bcb321c5da).


sherimickelson (approved):
Thanks @sjsprecious  for fixing this issue.  I think it looks good.

johnmauff (approved):

PR #40, opened 2022-03-22
Jmd fix gpu

johnmauff:
This update addresses several issues associated with inconsistent parameters between CPU and GPU.  We have also modified the Makefile and added a new file bcast.F that performs both an MPI reduction and a a OpenACC update device call.  This update also includes some fixes to the reduction calculations from the domaindiag.F.


src/param.F
```diff
@@ -273,7 +275,7 @@ subroutine param(namelist_input,                                  &
       call MPI_BCAST(idiss    ,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
       call MPI_BCAST(efall    ,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
       call MPI_BCAST(rterm    ,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
-      call MPI_BCAST(wbc      ,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
+      !KLUDGE call MPI_BCAST(wbc      ,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
```
sjsprecious, line 278: why is this line commented out?

src/param.F
```diff
@@ -220,12 +221,13 @@ subroutine param(namelist_input,                                  &
...
       close(unit=20)
-      !$acc enter data copyin(dx,dy,dz,wbc,ebc,sbc,nbc)
       
 
 #ifdef MPI
       endif  myid0
+#endif
 
+#ifdef MPI
       call MPI_BCAST(dx    ,1,MPI_REAL,0,MPI_COMM_WORLD,ierr)
```
sjsprecious, line 231: shall we change these calls to the new `bcast_int` or `bcast_real` wrappers as well?

sjsprecious (changes requested):

src/param.F
johnmauff: I comment out this particular call for the wbc variables because I added one at the end of the #ifdef MPI block.  So this was the test variable.

src/param.F
johnmauff: We will eventually replace all of these MPI_BCAST calls.    This will be a decent amount of work as there will be a need to add the corresponding $acc declare create() calls for all of the variables.  I thought that this was beyond the scope for this particular pull request.

src/param.F
sjsprecious: I see. Thanks!

src/param.F
sjsprecious: Thanks.

sherimickelson (approved):

sjsprecious (approved):

johnmauff:
Addresses MPI reduction, add broadcast wrappers, deactivated GPU-Direct

PR #41, opened 2022-03-30
Addressed bug in t2pcode.

johnmauff:
This fixes some an issue in the t2pcode subroutine.  Previous commits to the branch deactivated a loop on the GPU on t2pcode.  This reactivates this loop which has a iteration dependency by marking the loop as sequential.


src/Makefile
```diff
@@ -139,6 +139,7 @@ endif
 #-----------------------------------------------------------------------------
 ifeq (${FC},ifort)
     OPTS         += -g -O3 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback
+    #OPTS         += -g -O3 -xHost -ip -assume byterecl -fp-model precise -ftz -traceback
```
sjsprecious: Remove this commented flag.

src/misclibs.F
```diff
@@ -3013,12 +3013,14 @@ subroutine calccfl(nstat,rstat,dt,acfl,uh,vh,mh,ua,va,wa,writeit)
       mmax(2)=myid
       call MPI_ALLREDUCE(mmax,nmax,1,MPI_2REAL,MPI_MAXLOC,   &
                          MPI_COMM_WORLD,ierr)
+      !!$acc update device(mmax)
```
sjsprecious, line 3016: Looks like this directive is not used. Maybe remove it?

sjsprecious (changes requested):

src/misclibs.F
johnmauff: Actually, I want to keep these two commented-out update device calls.  The inclusion of this one and the one on line 3023 actually cause the GPU-enabled version of the code to blow up. I want to come back and understand the reason why.

johnmauff:
I have added several new namelist files to this PR that are scaled versions of already existing files.  Curiously, while some run successfully to completion others do not.  This is likely an indication that they trigger additional branches of the code that have existing bugs.  Here are the list and output of runs on a single v100 GPU:

namelist.baseline.dx=40m.short               Runs successfully
namelist.baseline.dx=10m.short-256        Runs successfully
namelist.baseline.dx=2.5m.short-128       NaNs from domaindiag, out of range numbers from statpack
namelist.baseline.dx=2.5m.short-160       NaNs from domaindiag, out of range numbers from statpack 
namelist.baseline.dx=2.5m.short-192       Runs successfully 
namelist.baseline.dx=2.5m.short-200      Misaligned address (may be due to memory issues)

src/misclibs.F
sjsprecious: Thanks for the explanation. After looking through these loops, I guess the issue may come from the index for the maximum value, as we discussed before that the GPU code may not return the last index for the maximum value as  the CPU code does. A similar issue exists in the maxmin.F as well. Shall we convert them to a sequential run in a separate PR or in this PR?

johnmauff: Let's make it for a separate PR.

src/Makefile
johnmauff: addressed

sjsprecious:
Thanks for this PR. @johnmauff . I tested the changes and it worked fine for both single GPU and MPI+GPU runs with `asd03-verify` namelist. However, with the `baseline.dx=40m` namelist, the single GPU run fails with an error related to CFL number.

sjsprecious (approved):

PR #42, opened 2022-04-18
Fix the GPU bugs in CM1 with updated droplet code

sjsprecious:
This pull request fixes some existing GPU bugs on the latest `gpu-opt` branch:

- `NaN` value in the output log
- Wrong `grdscl` and `min boundary-layer depth` in the output log
- Replace `acc declare copyin` by `acc declare create` and `acc update device` directives
- Wrong implementations for some `acc reduction` directives
- Replace `ieee_is_nan()` with a naive implementation to detect the `NaN` value (`ieee_is_nan()` does not work on GPU)
- Replace the `matmul` and `dot_product` with an explicit expression for small matrix/vector

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -74,7 +74,7 @@ endif
 #  NVHPC compiler
 #-----------------------------------------------------------------------------
 ifeq (${FC},nvfortran)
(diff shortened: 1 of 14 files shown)


johnmauff (approved):

sherimickelson (approved):
Thanks @sjsprecious for this fix!

PR #49, opened 2022-04-27
Open acc reduction cleanup

johnmauff:
This PR addresses several bugs associated with the placement of global integrals into GPU memory.  They were moved back to the CPU.  It eliminates a large number of '$acc compare' directives and some commented out stop statements used for debugging.  It did add a new routine used for debugging 'maxmin2dhalo' and added it to several locations in the code.

### DIFF
--- a/src/kessler.F
+++ b/src/kessler.F
@@ -389,6 +389,7 @@
       real :: esl,qvs,qvnew,qcnew,cvml,rm,lhv,tlast,dqv
       real :: converge,t1,d1,tem,ql,qi,rdt
       double precision :: dum
+      double precision :: bud1tmp,bud2tmp
       double precision, dimension(nk) :: bud1,bud2
       logical :: doit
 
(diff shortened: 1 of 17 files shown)


src/misclibs.F
```diff
@@ -4014,27 +4018,44 @@ subroutine calcmoe(nstat,rstat,ruh,rvh,rmh,rho,ua,va,wa,qv,ql,qi,vr)
...
+      call MPI_REDUCE(qltmp,var,1,MPI_DOUBLE_PRECISION,MPI_SUM,0,  &
+                      MPI_COMM_WORLD,ierr)
+      qltmp=var
+
+      var=0.0d0
+      call MPI_REDUCE(vrtmp,var,1,MPI_DOUBLE_PRECISION,MPI_SUM,0,  &
+                      MPI_COMM_WORLD,ierr)
+      vrtmp=var
+
       if(myid.eq.0)then
```
sjsprecious, line 4230: Are these for diagnostic only and will be removed later?

src/kessler.F
```diff
@@ -803,14 +813,16 @@ subroutine fallout_GPU(dt,train,ruh,rvh,zh,mh,mf,rain,prate,rr,rho,   &
       enddo
     enddo
     enddo
-
     !$acc end parallel
+
+    !JMD FIXME:  This loop looks incorrect. what is going on with the n-loop
```
sjsprecious, line 818: How do you know that this loop is incorrect?

src/misclibs.F
johnmauff: Yes, these are for diagnostic purposes only.  I wanted to identify which parts of the TMW variable was problematic.

src/kessler.F
johnmauff: This is a note to myself to come back and look at this code in more detail.  Note that the currently dx=40m namelist does not activate this code.

sjsprecious: Thanks for the explanation.

sjsprecious (approved):

PR #50, opened 2022-04-28
Fix some OpenACC reduction issue and unnecessary data movement for GPU

sjsprecious:
This PR fixes:
- some OpenACC reductions in the `get_avg_uvt` and `get_avg_uvtq` subroutines
- some unnecessary data movement in the `comm.F` code

This does not fix the boundary layer depth issue but reduces the error of `max w variance`.

### DIFF
--- a/src/comm.F
+++ b/src/comm.F
@@ -7530,7 +7530,7 @@
       ! send west
       if(ibw.eq.0)then
         !$omp parallel do default(shared) private(j)
-        !$acc parallel loop private(j)
+        !$acc parallel loop gang vector default(present) 
         do j=1,nj
           west(j)=s(1,j)
         enddo
@@ -7553,7 +7553,7 @@
       if(ibe.eq.0)then
 !$omp parallel do default(shared)   &
 !$omp private(j)
-        !$acc parallel loop private(j)
+        !$acc parallel loop gang vector default(present)
         do j=1,nj
           east(j)=s(ni,j)
         enddo
@@ -7576,7 +7576,7 @@
       if(ibn.eq.0)then
 !$omp parallel do default(shared)   &
 !$omp private(i)
-        !$acc parallel loop private(i)
+        !$acc parallel loop gang vector default(present)
         do i=1,ni
           north(i)=s(i,nj)
         enddo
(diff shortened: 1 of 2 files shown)


src/comm.F
```diff
@@ -13817,7 +13822,10 @@ subroutine prepcorners_GPU(s,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2,  &
 
       integer :: i,j
       logical, parameter :: Debug = .FALSE.
-      !$acc declare present(s)
+
+      !$acc data present(s,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2, &
+      !$acc              pw1,pw2,pe1,pe2,ps1,ps2,pn1,pn2)
+
```
johnmauff, line 13828: @sjsprecious, I don't understand the purpose of this set of directives for only a small region. If we want to make sure that data is present why are we not using a declare present() clause like is used throughout the rest of CM1.

sjsprecious: Thanks @johnmauff. I think it is equivalent to use `acc declare present (s,nw1,...)` here but it is just my personal preference to use `acc data present` directive. I used to use `acc declare` for module variables only, which helps me avoid some confusions.

johnmauff: I guess I have the opposite personal preference.  In particular, it requires an $acc end data which seems strange because the variable does not stop being present right.  It is a persistent variable on the device.  @sjsprecious Is there an issue with OpenMP offload syntax that would suggest we should use one versus the other?

sjsprecious: OpenMP offload syntax does not have a `present` clause right now. It uses `data map (to|from|alloc)` directive for data transfer, which is translated to `acc data copyin|copyout|create` directive in OpenACC. 
Regarding the `omp declare target` directive, there is no clause like `omp declare target create` in OpenMP compared to the `acc declare create` in OpenACC. Thus there is always data transfer when we use `omp declare target` directive, which could cause undesired effect.

sjsprecious: In addition, `$acc end data` here does not mean that the variables will not be present after this subroutine.
Adding `acc data` here is just a sanity check to me since sometimes `default(present)` clause is missing from the `acc parallel` directive and we may accidentally cause some unexpected data movement.

johnmauff: Ok, thanks for the clarification.

johnmauff (approved):

PR #51, opened 2022-05-07
Bcast update

johnmauff:
This PR replaces explicit MPI_BCAST calls with a simple wrapper bcast_{int,log,real}.  If the optional parameter device is set to be true, then it updates the GPU memory.  This update also moves some control flow variables that were not necessary to be GPU-resident back to the CPU.  The correctness stats for this PR is as follows:

CPU versus OpenACC:  27, 54, 7
four largest differences: KSVMAX, KSHMAX, TKEMAX, KHVMAX

CPU versus OpenACC + MPI: 30, 54, 4
four largest differences: KSVMAX, DIVMIN, DIVMAX, PPMAX


src/cm1.F
```diff
@@ -579,8 +578,8 @@ program cm1
       njp1 = nj+1
       nkp1 = nk+1
       nkm1 = nk-1
-      !$acc update device(nip1,njp1,nkp1,nkm1,ni,nj,nk,nparcels, &
```
sjsprecious, line 582: why is `nparcels` removed here?

src/cm1.F
```diff
@@ -2659,7 +2659,6 @@ program cm1
           call calccflquick(dt,uh,vh,mh,u3d,v3d,w3d,reqc)
 #ifdef MPI
           call mpi_wait(reqc,mpi_status_ignore,ierr)
-          !$acc update device(cflmax)
```
sjsprecious, line 2662: I think `cflmax` is calculated on the device. If so, removing this line is likely to report wrong results for the following print statement.

src/input.F
```diff
@@ -207,11 +207,11 @@ MODULE input
            base_pbot,base_ptop,base_thbot,base_thtop,base_qvbot,base_qvtop,   &
            base_tbot,base_ttop,base_pibot,base_pitop
 
-!$acc declare create(dx,dy,dz,dtl,timax,run_time,tapfrq,rstfrq,    &
-!$acc                statfrq,prclfrq,diagfrq,kdiff2,kdiff6,fcor,   &
+!$acc declare create(dx,dy,dz,dtl,timax,run_time, &
+!$acc                kdiff2,kdiff6,fcor,   &
 !$acc                alph,kdiv,wbc,ebc,sbc,nbc,rdx,rdy,rdz,minx,   &
-!$acc                maxx,miny,maxy,cflmax,ksmax,zt,rzt,umove,     &
-!$acc                vmove,viscosity,pr_num,sc_num,maxz,cgs1,cgs2, &
```
sjsprecious, line 214: I think these variables are used in some calculations on GPU. Why are they deleted here?

src/cm1.F
johnmauff: @sjsprecious  nparcels variable is removed because it is not needed to be on the device.  One of my goals in this PR was to only have variables present on the device if it was necessary.  Otherwise, I let the runtime move them over as needed.  I am concerned that having too many replicated variables present on both CPU and GPU will make the code more fragile than necessary.

src/cm1.F
johnmauff: @sjsprecious The scalar cflmax variable is the result of a reduction operator.  So once it is calculated on the GPU it should be updated on the CPU.

src/input.F
johnmauff: @sjsprecious during the creation of this PR I removed a large number of variables from GPU.  In this case, the cflmax, ksmax, umove, and vmove variables were not needed on the GPU.  I removed a large number of other control variables as well.  I consistently checked the correctness of each variable that I removed from the GPU.

src/cm1.F
johnmauff: @sjsprecious, I returned the nparcels variable to the GPU.  Instead of updating the device in the cm1.F file, I moved it back to the param() subroutine where it is initially set.

src/input.F
johnmauff: @sjsprecious I returned the umove and vmove variables to the GPU

src/input.F
```diff
@@ -161,16 +161,16 @@ MODULE input
               nutk,nvtk,nwtk
 
 !$acc declare create(nqs1,nqs2,nqr,nqi,nqs,nqg,nnci,nncc, &
-!$acc                ptype,psolver,ngxy,ngz,nx,ny,nz,axisymm, &
+!$acc                ptype,ngxy,ngz,nx,ny,nz, &
 !$acc                nparcels,npvals,npvars,ni,nj,nk,nip1,njp1, &
-!$acc                nkm1,nkp1,ib,ie,jb,je,kb,ke,terrain_flag, &
```
johnmauff, line 166: @sjsprecious It appears that axisymm and terrain_flag are also still needed on the GPU.

johnmauff: @sjsprecious I have returned several additional variables to the GPU.  I think all variables that I erroneously removed from the GPU are now back.

johnmauff:
@sjsprecious, I have updated the comment with the correctness verification statistics.

sjsprecious:
@johnmauff thanks for the updates. I checked out the `gpu-opt` branch and performed CPU and MPI+GPU tests with `tqnudge` turned off. The correctness verification statistics is:

CPU versus OpenACC + MPI: 27, 57, 4
four largest differences: KSVMAX, DIVMIN, DIVMAX, sv2.20

Thus I think the changes here improve the correctness.

sjsprecious (approved):

PR #53, opened 2022-05-13
These changes remove PCAST errors within the calcksmax subroutine.

sherimickelson:
Adding these changes removed the PCAST errors from the calcksmax subroutine.

### DIFF
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -3482,8 +3482,9 @@
 !JMD-FIXME
 !!$acc update host(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
 
-!$acc data copyout (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv) &
-!$acc      copyin (khh,vh,uh,mf)
+!!$acc data copyout (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv) &
+!!$acc      copyin (khh,vh,uh,mf)
+!$acc data create (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv,mmax,nmax)
 
       dtdx=dt*rdx*rdx
       dtdy=dt*rdy*rdy
@@ -3523,20 +3524,29 @@
             do i=1,ni
                tem = khh(i,j,k)*dtdx*uh(i)*uh(i)
                if( tem.eq.ksh(k) )then
+                 !$acc atomic write
                  imaxth(k)=i
+                 !$acc atomic write
                  jmaxth(k)=j
+                 !$acc atomic write
                  kmaxth(k)=k
                end if
                tem = khh(i,j,k)*dtdy*vh(j)*vh(j)
                if( tem.eq.ksh(k) )then
+                 !$acc atomic write
                  imaxth(k)=i
+                 !$acc atomic write
                  jmaxth(k)=j
+                 !$acc atomic write
                  kmaxth(k)=k
                end if
                tem = khv(i,j,k)*dtdz*mf(i,j,k)*mf(i,j,k)
                if( tem.eq.ksv(k) )then
+                 !$acc atomic write
                  imaxtv(k)=i
+                 !$acc atomic write
                  jmaxtv(k)=j
+                 !$acc atomic write
                  kmaxtv(k)=k
                endif
             end do
@@ -3551,18 +3561,26 @@
       imaxv=1
       jmaxv=1
       kmaxv=1
-      !$acc parallel loop gang vector default(present)
+      !$acc parallel loop default(present)
       do k=2,nk
          if(ksh(k).gt.fhmax)then
+           !$acc atomic write
            fhmax=ksh(k)
+           !$acc atomic write
            imaxh=imaxth(k)
+           !$acc atomic write
            jmaxh=jmaxth(k)
+           !$acc atomic write
            kmaxh=kmaxth(k)
          endif
          if(ksv(k).gt.fvmax)then
+           !$acc atomic write
            fvmax=ksv(k)
+           !$acc atomic write
            imaxv=imaxtv(k)
+           !$acc atomic write
            jmaxv=jmaxtv(k)
+           !$acc atomic write
            kmaxv=kmaxtv(k)
          endif
       enddo
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -3480,10 +3480,6 @@
       real maxksh,maxksv
 
 !JMD-FIXME
-!!$acc update host(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
-
-!!$acc data copyout (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv) &
-!!$acc      copyin (khh,vh,uh,mf)
 !$acc data create (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv,mmax,nmax)
 
       dtdx=dt*rdx*rdx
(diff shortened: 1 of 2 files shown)


johnmauff (commented):
@sherimickelson, this appears to use the old version of calccfl instead of the newly NVIDIA provided calccfl_fold version.  Sorry that I did not mentioned this previous.

src/misclibs.F
```diff
@@ -3482,8 +3482,9 @@ subroutine calcksmax(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
 !JMD-FIXME
 !!$acc update host(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
 
-!$acc data copyout (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv) &
-!$acc      copyin (khh,vh,uh,mf)
+!!$acc data copyout (ksh,ksv,imaxth,jmaxth,kmaxth,imaxtv,jmaxtv,kmaxtv) &
+!!$acc      copyin (khh,vh,uh,mf)
```
sjsprecious: remove these two lines if they are wrong.

src/misclibs.F
```diff
@@ -3523,20 +3524,29 @@ subroutine calcksmax(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
             do i=1,ni
                tem = khh(i,j,k)*dtdx*uh(i)*uh(i)
                if( tem.eq.ksh(k) )then
+                 !$acc atomic write
```
sjsprecious, line 3525: I am still confused about the behavior of `acc atomic write`. I would suggest using `loop seq` for the nested three loops to behave the same as the serial code first.

src/misclibs.F
```diff
@@ -3551,18 +3561,26 @@ subroutine calcksmax(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
       imaxv=1
       jmaxv=1
       kmaxv=1
-      !$acc parallel loop gang vector default(present)
+      !$acc parallel loop default(present)
```
sjsprecious: Change it to `loop seq` for the same reason above.

src/misclibs.F
sherimickelson: Thanks @sjsprecious , I removed these lines

src/misclibs.F
sherimickelson: @sjsprecious , I'm going to leave those lines in for the moment because as @johnmauff points out, this might not be an issue because the NVIDIA version does not use these.  Once we've verified the correctness of the new version, we can remove this code.

src/misclibs.F
sherimickelson: Thanks @sjsprecious , I made this change.

sherimickelson:
@sjsprecious and @johnmauff I created a new version of calcksmax and called it calcksmax_fold.  statpack.F calls this new version.  It doesn't seem to have changed any of the stat variable output and this still removed the PCAST error.

sjsprecious:
Hi @sherimickelson , thanks for this PR. I checked out the current `gpu-opt` branch and performed a CPU run. Then I checked out the `calcksmax_fix` branch and performed an MPI+GPU run. Both runs turned off the `tqnudge`. My results indicated that there were lot of differences in the stat output like the variables `KSHMAX` and `KSVMAX`.

sherimickelson:
@sjsprecious , thanks for checking for this.  My previous comment was in regards of the answers not being changed between [91846fb](https://github.com/george-bryan/CM1/pull/53/commits/91846fbfdbac673328615bdc09afbe6489070622) and [48d9d0e](https://github.com/george-bryan/CM1/pull/53/commits/48d9d0e035897642aa306d2dacf0fa183312643d).  

In regards to the answers being different between the CPU and GPU version, this is the tougher question to answer :).   I would expect that this changed answers because of the PCAST error that was coming out of it is now fixed and those are the two variables I modified.  Do you see anything incorrect in the code that would indicate that further modifications are needed?

src/misclibs.F
```diff
@@ -3630,6 +3644,210 @@ subroutine calcksmax(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
...
+      kmaxv=1
+
+!! Using a serial loop here so all the data is sync'd between consecutive kernel
+!calls.
+!! The RANGE = PAIRS or (PAIRS+1) as a way to manage odd numbers.
+      PAIRS=ni*nj/2           !! Round down.
+      RANGE=(ni*nj+1)/2       !! Round up.
+      do while ( RANGE > 1 )
+!! Now fold the 2-d array with half the threads each time.
+!$acc parallel loop gang vector private(i)
```
sjsprecious, line 3764: according to the comments, should `acc parallel loop seq` be used here?

src/misclibs.F
```diff
@@ -3630,6 +3644,210 @@ subroutine calcksmax(nstat,rstat,dt,uh,vh,mf,kmh,kmv,khh,khv)
...
+             ksv(i) = ksv(RANGE+1)
+             imaxtv(i) = imaxtv(RANGE+1)
+             jmaxtv(i) = jmaxtv(RANGE+1)
+             kmaxtv(i) = kmaxtv(RANGE+1)
+          endif
+        enddo
+        PAIRS=RANGE/2
+        RANGE=(RANGE+1)/2
+      enddo 
+
```
sjsprecious, line 3782: I think an `acc update host (ksh,imaxth,jmaxth,kmaxth,ksv,imaxtv,jmaxtv,kmaxtv)` directive is needed here.

sherimickelson: Thanks @sjsprecious for catching this.  I believe you're correct and I'll add the update host and verify correctness.

sherimickelson: calccfl_fold will need this change as well

sherimickelson: Both of these updates to the host were added in [e00ae25](https://github.com/george-bryan/CM1/pull/53/commits/e00ae254d365a73a0f2eaee2381c989be5fd04ff)

sherimickelson:
This latest commit ([e00ae25](https://github.com/george-bryan/CM1/pull/53/commits/e00ae254d365a73a0f2eaee2381c989be5fd04ff)) results in
62 stat variables show a mean relative difference > 1e-06
Top 4 variables are THPMAX, STHPMX, STHPMN, TMW

sjsprecious:
Thanks @sherimickelson for the update. The stat output looks worse compared to the current `gpu-opt` branch. Have you already merged the latest commit (https://github.com/george-bryan/CM1/commit/fd054dfb6648b9fe833df2d0435f12dca8c96044) in `gpu-opt` to your branch?

sherimickelson:
No, those stat numbers are without the merge.

sherimickelson:
@sjsprecious , with the merge, I'm now seeing
28 stat variables are identical.
54 stat variables show a mean relative difference > 1e-06

sjsprecious:
> @sjsprecious , with the merge, I'm now seeing 28 stat variables are identical. 54 stat variables show a mean relative difference > 1e-06 6 stat variables show a mean relative difference <= 1e-06

Thanks @sherimickelson. The new results show fewer variables that are either identical or larger than the tolerance but more variables that are smaller than the tolerance. To me this is fine unless we want to strictly keep the number of identical variables. If you have pushed all your changes to the `calcksmax_fix` branch, I would like to run it on my end and double check the results.

sherimickelson:
Thanks, @sjsprecious .  Yes, I've pushed all of my changes and that would be great if you can double check my results.

sjsprecious:
My results for verification runs:
CPU vs. OpenACC:
Identical variables: 30
variables with red_diff > 1.e-6: 53
variables with red_diff <= 1.e-6: 5
four variables with largest error: KSVMAX, KSHMAX, TMW, TKEMAX

CPU vs. MPI+OpenACC:
Identical variables: 27
variables with red_diff > 1.e-6: 58
variables with red_diff <= 1.e-6: 3
four variables with largest error: KSVMAX, KSHMAX, DIVMIN, TKEMAX

sjsprecious (approved):

PR #54, opened 2022-05-17
Addition of only clause for bc.F file.

johnmauff:
This PR adds an only clause to all uses of bc_module.  This allows the identification of routines within the bc_module that are no longer needed.   A large amount of the dead code in bc.F has been deleted resulting in the reduction of approximately 1K lines of code.  

CPU versus OpenACC: 27, 54, 7
four largest differences: KSVMAX, KSHMAX, TKEMAX, KHVMAX

CPU versus OpenACC + MPI: 27, 54, 5
four largest differences:  KSVMAX, KSHMAX, TKEMAX, KHVMAX


src/writeout.F
```diff
@@ -53,7 +53,7 @@ subroutine writeout(srec,urec,vrec,wrec,mrec,rtime,dt,fnum,nwrite,qname,qunit,
                         ! end_writeout
       use input
       use constants
-      use bc_module
+      use bc_module, only:
```
sjsprecious: can we just delete this line?

src/parcel.F
```diff
@@ -28,7 +28,7 @@ subroutine parcel_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
                                sigma,sigmaf,znt,rho,ua,va,wa,pdata)
       use input
       use constants
-      use bc_module
+      use bc_module, only:
```
sjsprecious: can we delete this line?

src/ib_module.F
```diff
@@ -714,7 +714,7 @@ subroutine ib_flx_init(bndy,hflxw,hflxe,hflxs,hflxn,dum1,dum2,     &
 
     use input
     use constants
-    use bc_module
+    use bc_module, only:
```
sjsprecious: can we delete this line?

src/droplet.F
```diff
@@ -30,7 +30,7 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
                                uten,vten,wten,thten,qten,pdata_locind)
       use input
       use constants
-      use bc_module
+      use bc_module, only:
```
sjsprecious: can we delete this line?

src/droplet.F
```diff
@@ -616,7 +616,7 @@ subroutine rk2_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,r
       !$acc routine seq
       use input
       use constants
-      use bc_module
+      use bc_module, only:
```
sjsprecious: can we delete this line?

src/droplet.F
```diff
@@ -933,7 +933,7 @@ subroutine BE_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,ru
...
```
sjsprecious: can we delete this line?

src/comm.F
```diff
@@ -14032,6 +14035,7 @@ subroutine prepcornert_GPU(t,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2,  &
 
       end subroutine prepcornert_GPU
 
+#if 0
```
sjsprecious: if the following subroutine is not used, I would suggest deleting it instead of using `#if 0`

src/comm.F
```diff
@@ -13874,14 +13875,15 @@ subroutine prepcorners_GPU(s,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2,  &
       end subroutine prepcorners_GPU
 
 
+#if 0
```
sjsprecious: if the following subroutine is not used, I would suggest deleting it instead of using #if 0

src/comm.F
```diff
@@ -13747,10 +13746,11 @@ end subroutine comm_all_s_GPU
 !cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 
 
+#if 0
```
sjsprecious: if the following subroutine is not used, I would suggest deleting it instead of using #if 0

src/bc.F
```diff
@@ -1,6 +1,23 @@
...
+  public :: ssopenbcw, ssopenbce, ssopenbcs, ssopenbcn, bcs_tend_halo
+
+  !CPU only 
+  public :: bcs, bcs2
+
+  !MPI + GPU enabled
+  public :: bcs2_GPU, bct2_GPU, bcu2_GPU, bcv2_GPU, bcw2_GPU, bcs2_2d_gpu
+
+  !MPI CPU only 
+  public :: bcs2_2d
```
sjsprecious, line 19: why the MPI CPU case could not use `bcs2_2d_gpu`?

src/azimavg.F
```diff
@@ -2745,7 +2745,7 @@ subroutine getpcent(xh,yh,pp3d,icenter,jcenter,xcenter,ycenter)
 
       use input
       use constants
-      use bc_module
+      use bc_module, only:
```
sjsprecious: could we delete this line?

src/bc.F
```diff
@@ -1,6 +1,23 @@
   MODULE bc_module
 
   implicit none
+  private
+
+  !GPU enabled
```
sjsprecious, line 6: why GPU and MPI+GPU need to call different subroutines here?

sjsprecious (changes requested):

src/writeout.F
johnmauff: Done

src/parcel.F
johnmauff: Done

src/ib_module.F
johnmauff: Done

src/droplet.F
johnmauff: Done

src/droplet.F
johnmauff: Done

src/comm.F
johnmauff: Yes, I have just deleted this code.

src/comm.F
johnmauff: Yes, I have deleted this code block.

src/azimavg.F
johnmauff: Done

src/bc.F
johnmauff: @sjsprecious, To aid debugging, I grouped those boundary update routines into those that were used in the serial code versus those that were used only when MPI was enabled.  Also, I grouped those boundary updates that were GPU enabled versus those that were not.

src/bc.F
johnmauff: @sjsprecious, Yes I could replace all uses of bcs2_2d with the bcs2_2d_GPU call.  However, I am a bit conflicted if I should minimize the number of changes or maximize the number of changes such that overall code volume is reduced.  I reverted the large number of changes in the bc.F subroutine because it was confusing during our code review yesterday.

sjsprecious: Thanks @johnmauff . That makes sense and it is better to do it in a separate PR.

src/bc.F
sjsprecious: Thanks @johnmauff for the explanation. Based on my understanding, using MPI and GPU or not only requires the activation of some CPP flags. Thus we should be able to use a single subroutine for different run types, which also reduces the risk of introducing inconsistency between different versions. But this may be a separate PR, too.

sjsprecious:
The changes here look good to me but I still do not understand why the stat output will change compared to the `gpu-opt` branch. I also do not understand why the total number of stat outputs will differ between different comparisons.

sjsprecious:
Since the metric output was revised, I used the `bc_cleanup` branch and generated the CPU and GPU runs for verification tests. Below are my results:

CPU vs. OpenACC:
Identical variables: 31
variables with red_diff > 1.e-6: 48
variables with red_diff <= 1.e-6: 4
four variables with largest error: KSVMAX, KHVMAX, KHHMAX, KSHMAX

CPU vs. MPI+OpenACC:
Identical variables: 31
variables with red_diff > 1.e-6: 48
variables with red_diff <= 1.e-6: 4
four variables with largest error: KSVMAX, KSHMAX, TKEMAX, KHVMAX

sjsprecious (approved):

PR #55, opened 2022-05-19
Clean up and improve coding-style of CM1 GPU code 

sjsprecious:
This PR mainly:
- Add `default(present)` clause to `acc parallel` and remove the `private` clause.
- Add missing `gang` or `vector` clause wherever applicable.
- Change the coding style of some OpenACC directives to improve readability.

I found a few GPU bugs during the code changes but I would report them in a separate PR.

I merged in the latest `gpu-opt` branch (https://github.com/george-bryan/CM1/commit/e5479c45e5281efcb126289e593fb040be65b619) and performed CPU (intel compiler), OpenACC and MPI+OpenACC runs. The verification output is summarized below:

CPU vs. OpenACC:
Identical variables: 31
variables with red_diff > 1.e-6: 48
variables with red_diff <= 1.e-6: 4
four variables with largest error: KSVMAX, KSHMAX, KHVMAX, KHHMAX

CPU vs. MPI+OpenACC:
Identical variables: 31
variables with red_diff > 1.e-6: 49
variables with red_diff <= 1.e-6: 3
four variables with largest error: KSVMAX, KSHMAX, TKEMAX, KHVMAX


sjsprecious:
My verification plot indicates that MPI+OpenACC has smaller error than OpenACC, inconsistent with the plot using the latest `gpu-opt` branch.

image

src/kessler.F
```diff
@@ -1015,36 +1015,25 @@ subroutine geterain_GPU(dt,cpx,lx1,erain,ruh,rvh,t,rr,q3d,vr)
       double precision :: budt
```
johnmauff, line 1015: I thought that we were not going to remove any B4B regions in this PR.

src/parcel.F
```diff
@@ -1153,48 +1146,34 @@ subroutine parcel_interp(dt,mtime,xh,uh,ruh,xf,uf,yh,vh,rvh,yf,vf, &
 
       !print *,'parcel_interp: point #12'
     IF( prqsl.ge.1 )THEN
-#ifdef _B4B02F
```
johnmauff, line 1156: Why are we removing B4B regions in this PR?

src/solve2.F
```diff
@@ -1733,22 +1733,14 @@ subroutine solve2(nstep,rbufsz,num_soil_layers,                  &
         !          conservation of total dry-air mass
         !FIXME
         !print *,'solve2: before dumk{1,2} loop: axisymm: ',axisymm
-#ifdef _B4B06R
-        !$acc update host(rho,ruh,rvh,rmh,pi0,pp3d,xf)
-#else
         !$acc parallel reduction(+:tmp1,tmp2)
```
johnmauff, line 1736: I would suggest that we do not inclose an if statement within a parallel region.  It would be easier to read if we moved the parallel region inside the if block.  Such a change would potentially eliminate the need for axisymm variable to be on the GPU.

src/kessler.F
sjsprecious: Actually I am not sure why we still want to keep the B4B regions in the CM1 code since we are not using them at all. Thus I prefer to clean them up while I am changing the code style.

src/parcel.F
sjsprecious: See the previous reply.

src/pdef.F
```diff
@@ -77,37 +75,40 @@ subroutine pdefx1(xh,arh1,arh2,uh,rho0,gz,rgz,rru,advx,dum,mass,s0,s,dt,flag,wes
...
             if(rru(ni+1,j,k).le.0.0)then
               dum(i,j,k)=rho0(i,j,k)*s0(i,j,k)
             endif
           enddo
           enddo
+          !$acc end parallel
         endif
+
         if(timestats.ge.1) time_pdef=time_pdef+mytime()
         !call flush(6)
```
johnmauff, line 110: Lets delete the flush(6) call

src/solve2.F
sjsprecious: I totally agree with this comment. But moving `acc parallel` into an `if` statement could be a NBFB change if the module variable is not set properly on the device. That is why I do not make changes in this PR.

src/pdef.F
```diff
@@ -341,36 +334,39 @@ subroutine pdefy1(vh,rho0,gz,rgz,rrv,advy,dum,mass,s0,s,dt,flag,south,newsouth,n
...
-          !$acc parallel loop gang collapse(2) private(i,k)
+          !$acc parallel loop gang vector collapse(2) default(present)
           do k=1,nk
           do i=1,ni
             if(rrv(i,nj+1,k).le.0.0)then
               dum(i,j,k)=rho0(i,j,k)*s0(i,j,k)
             endif
           enddo
           enddo
+          !$acc end parallel
```
johnmauff, line 364: @sjsprecious you are inconsistent in your addition of the '$acc end parallel' clause. Sometimes it is present while others it is not.

src/bc.F
```diff
@@ -2552,8 +2526,9 @@ subroutine restrict_openbc_we(rvh,rmh,rho0,u3d)
       !$acc declare present(rvh,rmh,rho0,u3d)
 
       !$acc data create(temout,temin)
+
       !$omp parallel do default(shared) private(k)
-      !$acc parallel loop gang vector private(k)
+      !$acc parallel loop gang vector default(present)
```
johnmauff, line 2531: Missing an outer reduction clause for temp1, and temp2

sjsprecious: I wonder if this is a NBFB change so I did not include it in this PR.

src/bc.F
```diff
@@ -2679,17 +2657,15 @@ subroutine restrict_openbc_sn(ruh,rmh,rho0,v3d)
       double precision :: fluxout,fluxin,tem,u1,t3
       double precision, dimension(nk) :: temout,temin
 
-!$omp parallel do default(shared)   &
-!$omp private(k)
+!$omp parallel do default(shared) private(k)
```
johnmauff, line 2660: @sjsprecious, This routine does not appear to be OpenACC enabled.  Have you noticed this?  Is it in other modifications?

src/domaindiag.F
```diff
@@ -10531,63 +10456,56 @@ subroutine scalar_flux(weps,c1,c2,mf,rf0,s   ,dum3,dum4,dum5,w3d,khv,sflux ,wavg
           if(bc_temp.eq.1)then
             ! specified theta at boundary
 
-#ifndef _B4B26R
-            !$acc loop vector collapse(2) reduction(+:sfst1,sfst2) 
-#endif
+            !$acc loop gang vector collapse(2) reduction(-:sfst1,sfst2) 
             do j=1,nj
             do i=1,ni
               sfst1 = sfst1-(viscosity/pr_num)*2.0*(savg(1)-ptc_bot)*rdz*mf(1,1,1)
```
johnmauff, line 10462: It appears that ptc_bot and ptc_top are not likely initialized properly on the GPU.  Is this addressed in a future PR?

johnmauff (commented):
@sjsprecious I was a bit surprised to see a large number of changes that removed the B4B regions.  They all look fine though.

src/kessler.F
johnmauff: OK, I am fine with that.

johnmauff:
@sjsprecious,   It would probably be a good idea to include a second verification plot in this PR.  The one that you included previously was before the merge of the getgamk bug fix.

src/bc.F
sjsprecious: Yes, I noticed this as well. I thought this was restricted to run on the CPU only like some subroutines in `adv.F` and thus I did not ever modify it.

sjsprecious:
> @sjsprecious I was a bit surprised to see a large number of changes that removed the B4B regions. They all look fine though.

Sorry about the large amount of elimination of BFB regions and the inconvenience it caused for reading. I think we eventually need to remove those regions and they are not used in our tests anyway. Thus I prefer to remove them while I am going through the codes.

src/domaindiag.F
sjsprecious: Yes, I guess it will be a NBFB change and I plan to fix it in a separate PR.

sjsprecious:
> @sjsprecious, It would probably be a good idea to include a second verification plot in this PR. The one that you included previously was before the merge of the getgamk bug fix.

@johnmauff, thanks for your suggestions. I merged your changes of fixing the `getgamk` bug and reran the verification tests (with `tqnudge = .true.`). The new results are listed below:

CPU vs. OpenACC:
Identical variables: 29
variables with red_diff > 1.e-6: 48
variables with red_diff <= 1.e-6: 6

CPU vs. MPI+OpenACC:
Identical variables: 26
variables with red_diff > 1.e-6: 57
variables with red_diff <= 1.e-6: 0

The results are identical with those in https://github.com/george-bryan/CM1/pull/56.

The new plot looks like:
image

src/solve2.F
johnmauff: Ok, let's move the placement of the 'acc parallel' directive on a future PR.

src/pdef.F
sjsprecious: Thanks. I found one missing directive and added it. Let me know if you still see more in the code.

src/domaindiag.F
johnmauff: Thanks

src/bc.F
johnmauff: This subroutine is called from the GPU region of the code.  So this is likey a bug.

src/pdef.F
johnmauff: ok

src/bc.F
johnmauff: ok

sherimickelson (approved):

PR #57, opened 2022-05-20
Added the only clause to the input module in several acoustic solvers

johnmauff:
This PR adds the use of the FORTRAN only clause for the 'input' module in several acoustic solvers.  This change has allowed for the identification of two module variables (smeps, csound) that were potentially not set properly on the GPU.  While it does not appear to change answers.  

CPU versus GPU
54 stat variables show a mean relative difference > 1e-06

CPU versus GPU+MPI

### DIFF
--- a/src/anelp.F
+++ b/src/anelp.F
@@ -19,7 +19,11 @@
                        wa,w3d,wten,                                 &
                        ppi,pp3d,phi1,phi2,cfb,cfa,cfc,              &
                         d1, d2,pdt,lgbth,lgbph,rhs,trans,dttmp,nrk,rtime,mtime)
-      use input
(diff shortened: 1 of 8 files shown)


sjsprecious:
This PR looks straightforward to me. One question: there is no `acc routine seq` in the revised subroutine but why adding `only` clause is able to detect the missing module variables on the device?

johnmauff:
I actually found the missing module variables on the device using old fashion code inspection.  I added the only clause which generated missing symbol error messages.  Based on this list of module variables that the subroutine used, I looked through the code to see if any of them were located in parallel loops.  I ended up finding two that were indeed in parallel loops.

sjsprecious (approved):

johnmauff:
@sherimickelson,  i just added you as a reviewer for this PR.

sherimickelson (approved):

PR #59, opened 2022-05-26
Add missing OpenACC reduction clause and improve coding style

sjsprecious:
This PR further improves some coding styles and fixies  some reduction clauses for the GPU code.

Verification results:

CPU versus OpenACC GPU
four variables with maximum error: KSVMAX, KSHMAX, TMW, TKEMAX

CPU versus OpenACC GPU + MPI
four variables with maximum error: STHPMX, THPMAX, TMW, PPMAX

This is consistent with the current `gpu-opt` branch.

### DIFF
--- a/src/domaindiag.F
+++ b/src/domaindiag.F
@@ -10102,25 +10102,18 @@
 
   !----------------------
 
-    
-#ifndef _B4B21R
-    !$acc parallel default(present) private(i,j,k) reduction(+:vbart) reduction(+:vfrt,vfdt)
+    !$acc parallel default(present)
     !$acc loop gang vector
-#endif
     do k=1,nk
       vbar(k) = 0.0
     enddo
+    !$acc end parallel
 
-#ifdef _B4B21R
-    !$acc update host(dum3)
-#else
+    !$acc parallel default(present) reduction(+:vbart)
     !$acc loop gang
-#endif
     do k=2,nk
       vbart = 0.0d0
-#ifndef _B4B21R
       !$acc loop vector collapse(2) reduction(+:vbart)  
-#endif
       do j=1,nj
       do i=1,ni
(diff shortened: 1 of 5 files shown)


src/domaindiag.F
```diff
@@ -10102,39 +10102,26 @@ subroutine momentum_flux(u3d,v3d,w3d,t13,t23,m13,m23,rf,rf0,c1,c2,dum3,dum4,dum5
 
   !----------------------
 
-    
-#ifndef _B4B21R
-    !$acc parallel default(present) private(i,j,k) reduction(+:vbart) reduction(+:vfrt,vfdt)
+    !$acc parallel default(present)
```
johnmauff, line 10103: @sjsprecious, you should be aware that the momentum_flux() subroutine is not currently called in CM1.  I had commented out its call because it cause an OpenACC error.  Unless you have already reactivated the code, I would encourage you to turn it back on to verify that it now works with your changes.

johnmauff (changes requested):
While I am hopeful that your changes will fix the bug in momentum_flux() it would be good to verify.

sjsprecious:
@johnmauff thanks for your comment. I did not realize that `momentum_flux()` was not activated before. I turned it on and reran the CPU and GPU tests. All the tests finished successfully and the verification stats remained the same. Should I remove the `#if 0` in the domaindiag.F for the `momentum_flux()` then?

johnmauff:
@sjsprecious, Thanks for fixing this bug.  Please go ahead and remove the #if 0 in the domaindiag.F code.

sjsprecious:
> @sjsprecious, Thanks for fixing this bug. Please go ahead and remove the #if 0 in the domaindiag.F code.

Thanks. I just removed it.

johnmauff (approved):

PR #62, opened 2022-06-01
Input only mods #4

johnmauff:
Yet another update that adds the only clause to the input module.  This has the same correctness statistics to the existing gpu-opt branch:

CPU Intel compared to GPU OpenACC

CPU Intel compared to GPU OpenACC+MPI 

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -1550,7 +1550,6 @@
       subroutine find_horizontal_location_index (loc_ind, loc, lb, ub, dsize, end_ind, ind)
       !$acc routine seq
 
-      use input, only :
       use constants
 
       implicit none
@@ -1634,7 +1633,6 @@
       subroutine find_vertical_location_index (loc_ind, loc, lb, ub, dsize, ind, is_ge)
       !$acc routine seq
 
-      use input, only :
       use constants
 
       implicit none
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -28,7 +28,11 @@
                                sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,             &
                                n3w1,n3w2,n3e1,n3e2,s3w1,s3w2,s3e1,s3e2,reqs_s,      &
                                uten,vten,wten,thten,qten,pdata_locind)
-      use input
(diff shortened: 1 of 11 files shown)


src/droplet.F
```diff
@@ -1539,7 +1550,7 @@ end subroutine interpolate_to_parcel
       subroutine find_horizontal_location_index (loc_ind, loc, lb, ub, dsize, end_ind, ind)
       !$acc routine seq
 
-      use input
+      use input, only :
```
sjsprecious: It seems that we do not need to use the `input` module here.

src/droplet.F
```diff
@@ -1623,7 +1634,7 @@ end subroutine find_horizontal_location_index
       subroutine find_vertical_location_index (loc_ind, loc, lb, ub, dsize, ind, is_ge)
       !$acc routine seq
 
-      use input
+      use input, only :
```
sjsprecious: It seems that we do not need to use the input module here.

src/ib_module.F
```diff
@@ -230,6 +232,7 @@ subroutine ib_init(xh,yh,xf,yf,sigma,sigmaf,zs,zh,bndy,kbdy,out3d,  &
         kmaxib = max( kmaxib , kbdy(i,j) )
       enddo
       enddo
+      !$acc update device(kmaxib)
```
sjsprecious: should this line be moved to after the `MPI_ALLREDUCE` call?

sjsprecious (changes requested):

src/droplet.F
johnmauff: Removed the use input statement.

src/droplet.F
johnmauff: Removed the use input statement.

src/ib_module.F
johnmauff: Thanks, @sjsprecious, this was a good catch.  I have moved this update device call to after the MPI_ALLREDUCE.

sjsprecious: Thanks @johnmauff . And unfortunately I need to remove this statement eventually after the discussion yesterday.

sjsprecious (approved):

PR #63, opened 2022-06-05
Add missing openacc directive in the restrict_openbc_sn subroutine

sjsprecious:
This PR adds the missing OpenACC directive in the `restrict_openbc_sn` subroutine in the `bc.F` code.

However, this subroutine is not called in the `namelist.baseline.dx=40m.short` because `roflux=0`.

The verification result is the same as the current `gpu-opt` branch:

### DIFF
--- a/src/bc.F
+++ b/src/bc.F
@@ -2680,44 +2680,56 @@
       real, dimension(ib:ie,jb:je+1,kb:ke) :: v3d
 
       integer i,j,k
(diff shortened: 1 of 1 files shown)


johnmauff (approved):
@sjsprecious, Thanks for fixing this bug, it looks good to me.  It is unfortunate that it does not address the timestep = 26 bug that we have been chasing.

sherimickelson (approved):

PR #64, opened 2022-06-06
Eliminated dead code in comm.F and bc.F

johnmauff:
This PR eliminates a large number of subroutines in bc.F and comm.F that are no longer used.  Specifically, these are CPU-only routines, that have been replaced by GPU verions.  The correctness statistics are unchanged.  

Intel compared to OpenACC+MPI

Intel compared to OpenACC


sherimickelson (approved):

sjsprecious:
As John reported, somehow we could not see the detailed changes in `comm.F` through the `Files changed`.

However, the `git diff` command line still worked. I could confirm that John just deleted the subroutines in `comm.F` that ran only on CPU and thus were no longer needed with the existence of corresponding GPU versions.

sjsprecious (approved):

PR #65, opened 2022-06-07
remove unnecessary module variables from acc directives

sjsprecious:
This PR removes some module variables not called by an ACC routine from the OpenACC directives (`acc declare create` and `acc update device`). Ideally it should not change the results if we update all the variables correctly on the device.

The new verification result indicates that the correctness is unchanged.


johnmauff (approved):
@sjsprecious Thanks for removing these variables from the device.  These changes looks good to me.

sherimickelson (approved):

PR #67, opened 2022-06-22
Fix gpu bugs and update Makefile

sjsprecious:
This PR:
- fixes some bugs for CM1 GPU code when working on a 4-D array
- adds some missing OpenACC directives
- updates Makefile with managed memory option for GPU runs and ability to apply different compiler options to different source codes
- improves some GPU coding styles

---

The new verification result indicates that the correctness is **unchanged** for GPU runs without managed memory.

---

The new verification result indicates that the correctness is **improved** for GPU runs with managed memory.

### DIFF
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -875,8 +875,8 @@
     ENDDO
 
     IF( nql1.gt.0 )THEN
-    !$acc parallel default(present) private(i,j,k,n)
     DO n=nql1,nql2
+      !$acc parallel default(present)
       !$acc loop gang vector collapse(3)
       do k=1,nk
         do j=1,nj
@@ -885,13 +885,13 @@
         enddo
         enddo
       enddo
+      !$acc end parallel
     ENDDO
-    !$acc end parallel
     ENDIF
 
       IF(iice.eq.1)THEN
-      !$acc parallel default(present) private(i,j,k,n)
       DO n=nqs1,nqs2
+        !$acc parallel default(present)
         !$acc loop gang vector collapse(3)
         do k=1,nk
           do j=1,nj
@@ -900,8 +900,8 @@
           enddo
           enddo
         enddo
+        !$acc end parallel
       ENDDO
-      !$acc end parallel
       ENDIF
 
       if(timestats.ge.1) time_misc02=time_misc02+mytime()
@@ -1014,8 +1014,8 @@
       ENDDO
       IF( nql1.gt.0 )THEN
         !$omp parallel do default(shared) private(i,j,k,n)
-        !$acc parallel default(present) private(i,j,k,n)
         do n=nql1,nql2
+        !$acc parallel default(present)
         !$acc loop gang vector collapse(3)
         DO k=1,nk
           do j=1,nj
@@ -1024,13 +1024,13 @@
           enddo
           enddo
         ENDDO
-        enddo
         !$acc end parallel
+        enddo
       ENDIF
       IF(iice.eq.1)THEN
-        !$acc parallel default(present) private(i,j,k,n)
         !$omp parallel do default(shared) private(i,j,k,n)
         do n=nqs1,nqs2
+        !$acc parallel default(present)
         !$acc loop gang vector collapse(3)
         DO k=1,nk
           do j=1,nj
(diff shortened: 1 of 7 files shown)


sherimickelson (approved):

src/misclibs.F
```diff
@@ -3930,8 +3930,9 @@ subroutine getrmw(nstat,rstat,xh,zh,ua,va)
       !FIXME
 !$omp parallel do default(shared)  &
 !$omp private(i,k,wspd)
-!$acc parallel loop gang vector collapse(2) default(present) private(i,k)
+!$acc parallel loop gang vector default(present)
       do k=1,nk
+        !$acc loop seq
```
johnmauff, line 3935: @sjsprecious You appear to be using two different directives to indicate a serial loop.  In this case $acc loop seq and in other places $acc serial.  Why are you being inconsistent?

src/misclibs.F
```diff
@@ -875,8 +875,8 @@ subroutine getqli(q,ql,qi)
     ENDDO
 
     IF( nql1.gt.0 )THEN
-    !$acc parallel default(present) private(i,j,k,n)
     DO n=nql1,nql2
+      !$acc parallel default(present)
```
johnmauff, line 879: @sjsprecious Not clear that this change does anything other than slow the code down.  Why are you launching multiple parallel regions instead of a single one?

src/solve1.F
```diff
@@ -399,7 +399,8 @@ subroutine solve1(nstep,rbufsz,num_soil_layers,                  &
       !$acc present(qa,qten) &
       !$acc present(kmh,kmv,khh,khv,tkea,tketen,nm,dissten,epst) &
       !$acc present(pta,ptten,u2pt,v2pt) &
-      !$acc present(tdiag,qdiag,udiag,vdiag,wdiag,kdiag)
+      !$acc present(tdiag,qdiag,udiag,vdiag,wdiag,kdiag) &
+      !$acc present(uavg,vavg,thavg,qavg,cavg)
```
johnmauff, line 403: @sjsprecious Why were these arrays added into a present clause here and not in other subroutines where they were also used like sfc_and_turb

src/testcase_simple_phys.F
```diff
@@ -453,8 +456,9 @@ subroutine simplerad(rho0,zh,zf,dum1,dum2,qt  ,ql  ,qa,frad,zir)
         frad(i,j,k) = fr1+fr2+fr3
       enddo
       enddo
-      enddo
       !$acc end parallel
+      enddo
+
       stop 'simplerad: at the end of the subroutine'
```
johnmauff, line 462: Probably can remove the stop statement from this subroutine.

johnmauff (commented):
@sjsprecious It is not clear to me why you changed every instance of the do n=nqs1,nqs2 loop.  While I suspect you addressed one bug, it is not clear to me if a large number of your changes had any impact to the code other than to slow it down by forcing the launch of (nqs2-nqs1) more parallel regions.

src/misclibs.F
sjsprecious: Thanks @johnmauff for your comment. I think this change is needed to get the correct answer. Otherwise, putting `acc parallel` at the `n` loop level would not perform the `ql` calculation correctly IMO.

src/solve1.F
sjsprecious: I agree that this addition seems unnecessary and I will remove it. Ideally I want to remove all these `present` directives after we add `default(present)` to all the `acc parallel` directives eventually. But this may take a while.

src/misclibs.F
sjsprecious: @johnmauff `acc serial` directive will run the whole compute kernel in serial, while `acc loop seq` directive can run a specific loop in serial, which is the case here.

src/testcase_simple_phys.F
sjsprecious: Sure and I think this subroutine is not called in our current test.

src/misclibs.F
johnmauff: Thanks for the clarification.

src/misclibs.F
johnmauff: @sjsprecious, still not convinced of your explanation.  However, I see no reason to hold up the commit over it.

johnmauff: @sjsprecious after you explanation this makes sense the reason why you relocated the n-loop

PR #68, opened 2022-07-13
Bring a deterministic random number generator from CESM to CM1

sjsprecious:
This PR introduces a deterministic random number generator from CESM to CM1.

Using this deterministic random number generator, the generated random number is independent of compiler types and number of MPI ranks in a particular run.

Below is an example comparison of the 0th-order droplet diagnostic between two CPU runs with different compilers and between CPU and MPI+GPU runs, demonstrating how this new random number generator reduces the discrepancy to a much smaller magnitude.

**CPU runs: intel vs. nvhpc**
image

**CPU run (intel) vs. MPI+GPU (nvhpc)**
image

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -203,6 +203,7 @@ $(info $$DM is [${DM}])
 #-----------------------------------------------------------------------------
 
 SRC   = constants.F \
+	mersennetwister_mod.F \
 	input.F \
 	adv.F \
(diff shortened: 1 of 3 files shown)


johnmauff (approved):
Thanks for this addition @sjsprecious.  It looks good to me.

sherimickelson (approved):

PR #69, opened 2022-07-14
Fix GPU bugs for the partnum calculation

sjsprecious:
This PR adds a missing OpenACC reduction clause for the calculation of `partnum` in the droplet diagnostic subroutine.

With this fix, I am able to obtain consistent 1st-order droplet diagnostic results between CPU and GPU runs.

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -1671,7 +1671,7 @@
       end do
       !$acc end parallel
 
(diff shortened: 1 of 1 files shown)


johnmauff (approved):
Great to see that you found another bug.

sherimickelson (approved):
@sjsprecious good find!  Thanks for fixing this so quickly!

PR #70, opened 2022-08-02
Enable the deterministic random number generator on GPU

sjsprecious:
This PR enables the deterministic random number on GPU by adding the OpenACC directives.

The GPU-enabled deterministic random number is later used in the droplet subroutine for re-injection of fallout droplets.

The results indicate that:
- the 1st-order droplet diagnostic is consistent between CPU and GPU runs.
- the 0th-order droplet diagnostic differs at a small magnitude of relative error.

**CPU runs: intel vs. openacc**
image

CPU run (intel) vs. MPI+GPU (nvhpc)
image

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -296,7 +296,11 @@ OBJS_M = $(addsuffix .o, $(basename $(SRC_M)))
 OBJS_OPENACC = $(addsuffix .o, $(basename $(SRC_NVTX)))
 
 FFLAGS  = $(OPTS)
-FFLAGS_M  = $(OPTS_M)
+ifeq (${USE_OPENACC_L},true)
+FFLAGS_M = $(OPTS_M) -L/glade/u/apps/opt/nvhpc/22.2/Linux_x86_64/22.2/cuda/11.6/lib64 -lcudart
+else
+FFLAGS_M = $(OPTS)
+endif
 AR      = ar cru
 
 .SUFFIXES:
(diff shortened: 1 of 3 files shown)


src/Makefile
```diff
@@ -296,7 +296,11 @@ OBJS_M = $(addsuffix .o, $(basename $(SRC_M)))
 OBJS_OPENACC = $(addsuffix .o, $(basename $(SRC_NVTX)))
 
 FFLAGS  = $(OPTS)
-FFLAGS_M  = $(OPTS_M)
+ifeq (${USE_OPENACC_L},true)
+FFLAGS_M = $(OPTS_M) -L/glade/u/apps/opt/nvhpc/22.2/Linux_x86_64/22.2/cuda/11.6/lib64 -lcudart
```
johnmauff: @sjsprecious So this change to the link line was due to adding a CUDA-based memory measurement right? It probably will not be used very frequently.  Maybe we should switch off this link option by default?

sherimickelson (approved):
@sjsprecious Thanks for adding this in.  It'll really help with verifying our future modifications.

src/Makefile
johnmauff: Thanks @sjsprecious, this PR now looks good.

sjsprecious: Thanks @johnmauff. I have commented out the lib path as you suggested.

johnmauff (approved):

PR #71, opened 2022-08-03
Fix a GPU error for find_vertical_location_index subroutine

sjsprecious:
We recently realize that when we work on a large domain with smaller grid size (e.g., dx=2.5m) and a large number of droplets (nparcels > 2,000), the GPU run of CM1 will crash by invoking the `find_vertical_location_index` subroutine.

We find a temporary solution by generating a local 1-D array `zf_tmp` for each CUDA thread and copying the information from `zf` array. This fixes the GPU error but also increases the CUDA memory usage.

The verification results using `namelist.baseline.dx=40m.short` indicate that:
- the 1st-order droplet diagnostic is consistent between CPU and GPU runs.
- the 0th-order droplet diagnostic differs at a small magnitude of relative error.

**CPU runs: intel vs. OpenACC**
image

**CPU run (intel) vs. MPI+GPU (nvhpc)**
image

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -103,7 +103,8 @@
       integer :: idropmethod
 
       logical, parameter :: debug = .false.
-      
+     
+      real, dimension(kb:ke+1) :: zf_tmp 
       type(randomNumberSequence) :: randomNumbers
 
 #ifdef _VERIFY_FIND_LOC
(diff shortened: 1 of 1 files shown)


sherimickelson (approved):
Thanks, @sjsprecious

johnmauff (approved):
Looks good to me.

PR #72, opened 2022-08-04
Eliminate remaining B4B blocks

johnmauff:
This pull request eliminates the remaining B4B blocks that are present in the code.  Using a NVHPC+MPI on CPU as a base, all of the diagnostics match with the exception of the following.

NVHPC+MPI on GPU
================
Original:
--------
46 stat variables are identical.
25 stat variables show a mean relative difference > 1e-06
12 stat variables show a mean relative difference <= 1e-06
Biggest differences:  KSVMAX,KSHMAX,VOR1KM,VOR2KM

modified code run
------------------
43 stat variables are identical.
25 stat variables show a mean relative difference > 1e-06
15 stat variables show a mean relative difference <= 1e-06

### DIFF
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -4144,13 +4144,14 @@
         foo4(k) = foo4t
       enddo
       !$acc end parallel
-      !$acc parallel loop vector default(present) reduction(+:ek,ei,ep,le)
+      !$acc parallel loop gang vector default(present) reduction(+:ek,ei,ep,le)
       do k=1,nk
         ek=ek+foo1(k)
         ei=ei+foo2(k)
         ep=ep+foo3(k)
         le=le+foo4(k)
       enddo
+      !$acc end parallel
       !$acc end data
  
       ek=ek*(dx*dy*dz)
@@ -4372,7 +4373,7 @@
       !$acc end parallel
       tmfu=0.0d0
       tmfd=0.0d0
-      !$acc parallel loop vector default(present) reduction(+:tmfu,tmfd)
+      !$acc parallel loop vector gang default(present) reduction(+:tmfu,tmfd)
       do k=1,nk
         tmfu=tmfu+foo1(k)
         tmfd=tmfd+foo2(k)
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -1,9 +1,7 @@
 #ifdef _B4B
 
-#undef _B4B01R
 
 !JMD-expensive
-#define _B4B02R
 #define _B4B04R
 #define _B4B05R
 #define _B4B06R
@@ -1155,41 +1153,28 @@
       integer i,j,k
       double precision :: t1,t2,t3
       double precision :: a1,a2,tem
-      double precision, dimension(nj) :: budj
-      double precision, dimension(nk) :: budk
       double precision :: budt
       !$acc declare present(ruh,rvh,rmh,rho,q3d)
 
 !----------------------------------------------------------------------
 
       !print *,'pdefq_GPU: pdscheme: ',pdscheme
-#ifndef _B4B01R
-      !$acc data create(budj,budk)
-#endif
+      !!$acc data create(budj,budk)
       !tem = dx*dy*dz
       IF(pdscheme.eq.1)THEN
 
-#ifdef _B4B01R
-        !$acc update host(asq,ruh,rvh,rmh,rho,q3d)
-#else
-        !!$acc parallel loop gang vector default(present) private(i,j,k)
-        !$acc parallel default(present) private(i,j,k) &
-        !$acc reduction(+:a1,t1,t2) reduction(+:a2) reduction(+:budt)
-        !$acc loop gang
-#endif
         !$omp parallel do default(shared) private(i,j,k,t1,t2,t3,a1,a2)
+        !$acc parallel default(present) private(i,j,k) &
+        !$acc reduction(+:a1,t1,t2) reduction(+:a2) reduction(+:budt,asq)
+        !$acc loop gang
         do j=1,nj
           budt = 0.0d0
-#ifndef _B4B01R
           !$acc loop vector reduction(+:budt)
-#endif
           do i=1,ni
             t1=0.0d0
             t2=0.0d0
             a1=0.0d0
-#ifndef _B4B01R
             !$acc loop vector reduction(+:a1,t1,t2)
-#endif
             do k=1,nk
               t1=t1+rho(i,j,k)*q3d(i,j,k)
               a1=a1+rho(i,j,k)*q3d(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
@@ -1200,33 +1185,22 @@
             t3=(t1+1.0d-20)/(t2+1.0d-20)
             if(t3.lt.0.0) t3=1.0d0
             a2=0.0d0
-#ifndef _B4B01R
             !$acc loop vector reduction(+:a2)
-#endif
             do k=1,nk
               q3d(i,j,k)=t3*q3d(i,j,k)
               a2=a2+rho(i,j,k)*q3d(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
             enddo
             budt = budt+a2-a1
           enddo
-          budj(j)=budt
+          asq=asq+budt*dx*dy*dz
         enddo
-
-#ifndef _B4B01R
         !$acc end parallel
-        !$acc parallel loop vector reduction(+:asq)
-#endif
-        do j=1,nj
-          asq=asq+budj(j)*dx*dy*dz
-        enddo
 
       ELSE
 
         !$omp parallel do default(shared) private(i,j,k,a1,a2)
-#ifndef _B4B01R
         !$acc parallel default(present) private(i,j,k,a1,a2) reduction(+:budt,asq)
         !$acc loop gang
-#endif
         do k=1,nk
           budt=0.0d0
           !$acc loop vector collapse(2) reduction(+:budt)
@@ -1239,25 +1213,11 @@
             budt=budt+a2-a1
           enddo
           enddo
-         budk(k)=budt 
+          asq=asq+budt*dx*dy*dz
         enddo
-
-#ifndef _B4B01R
-        !$acc loop vector reduction(+:asq)
-#endif
-        do k=1,nk
-          asq=asq+budk(k)*dx*dy*dz
-        enddo
-#ifndef _B4B01R
       !$acc end parallel
-#endif
 
       ENDIF
-#ifdef _B4B01R
-      !$acc update device(asq,q3d)
-#else
-      !$acc end data
-#endif
 
 !----------------------------------------------------------------------
 
@@ -1290,7 +1250,6 @@
 !----------------------------------------------------------------------
 
       !print *,'pdefq2_GPU: pdscheme: ',pdscheme
-!      !$acc data create(budj,budk)
       !tem = dx*dy*dz
       IF(pdscheme.eq.1)THEN
 
@@ -1319,7 +1278,6 @@
               q3d(i,j,k)=t3*q3d(i,j,k)
               a2=a2+rho(i,j,k)*q3d(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
             enddo
-            !budt = budt+a2-a1
             asq=asq+(a2-a1)*dx*dy*dz
           enddo
         enddo
@@ -1342,18 +1300,11 @@
             budt=budt+a2-a1
           enddo
           enddo
-          !budk(k)=budt 
           asq=asq+budt*dx*dy*dz
         enddo
-
-        !!$acc loop vector reduction(+:asq)
-        !do k=1,nk
-        !  asq=asq+budk(k)*dx*dy*dz
-        !enddo
         !$acc end parallel
 
       ENDIF
-!      !$acc end data
 
 !----------------------------------------------------------------------
 
@@ -1453,19 +1404,12 @@
         enddo
         !$acc end parallel
 
-#ifdef _B4B02R
-        !$acc update host(dum2d5,dum2d2) 
-#else
         !$acc parallel loop vector collapse(2) default(present) private(i,j) reduction(+:asq)
-#endif
         do j=1,nj
         do i=1,ni
           asq=asq+(dum2d5(i,j)-dum2d2(i,j))*tem
         enddo
         enddo
-#ifdef _B4B02R
-        !$acc update device(asq)
-#endif
 
       ELSE
 
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -4109,7 +4109,7 @@
       !$omp parallel do default(shared)  &
       !$omp private(i,j,k,u,v,w,tmp,qtot,tem)
       !$acc parallel default(present) private(i,j,k,u,v,w,tmp,qtot,tem) &
-      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t,ek,ei,ep,le)
+      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t)
       !JMD internal compiler error
       !!$acc loop gang reduction(+:ek,ei,ep,le)
       !$acc loop gang
@@ -4143,14 +4143,14 @@
         foo3(k) = foo3t
         foo4(k) = foo4t
       enddo
-      !$acc loop vector reduction(+:ek,ei,ep,le)
+      !$acc end parallel
+      !$acc parallel loop vector default(present) reduction(+:ek,ei,ep,le)
       do k=1,nk
         ek=ek+foo1(k)
         ei=ei+foo2(k)
         ep=ep+foo3(k)
         le=le+foo4(k)
       enddo
-      !$acc end parallel
       !$acc end data
  
       ek=ek*(dx*dy*dz)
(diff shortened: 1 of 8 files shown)


src/sfcphys.F
```diff
@@ -497,36 +492,18 @@ subroutine sfcflux(dt,ruh,xf,rvh,pi0s,ch,cq,pi0,thv0,th0,rf0,tsk,thflux,qvflux,m
     IF(imoist.eq.1)THEN
       tem = dt*dx*dy
       ! some budget calculations:
-#ifdef _B4B06R
-      !$acc update host(qvflux,ruh,rvh,rf) 
-#else
-      !$acc data create(bud1)
+      !$omp parallel do default(shared) private(i,j)
       !$acc parallel default(present) private(i,j) reduction(+:budtmp,qbsfc) 
       !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:qbsfc)` here?

src/sfcphys.F
```diff
@@ -336,7 +331,7 @@ subroutine sfcflux(dt,ruh,xf,rvh,pi0s,ch,cq,pi0,thv0,th0,rf0,tsk,thflux,qvflux,m
 
       integer :: i,j
       real :: pisfc,qvsat,tem,shf
-      double precision, dimension(nj) :: bud1
+      !double precision, dimension(nj) :: bud1
```
sjsprecious: delete it.

src/misclibs.F
```diff
@@ -1155,41 +1141,27 @@ subroutine pdefq_GPU(rmax,asq,ruh,rvh,rmh,rho,q3d)
...
 
-#ifdef _B4B01R
-        !$acc update host(asq,ruh,rvh,rmh,rho,q3d)
-#else
-        !!$acc parallel loop gang vector default(present) private(i,j,k)
+        !$omp parallel do default(shared) private(i,j,k,t1,t2,t3,a1,a2)
         !$acc parallel default(present) private(i,j,k) &
-        !$acc reduction(+:a1,t1,t2) reduction(+:a2) reduction(+:budt)
+        !$acc reduction(+:a1,t1,t2) reduction(+:a2) reduction(+:budt,asq)
         !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:asq)` here?

src/misclibs.F
```diff
@@ -1453,30 +1388,18 @@ subroutine pdefqtest(rmax,asq,ruh,rvh,rmh,rho,q3d,dum2d1,dum2d2,dum2d3,dum2d4,du
...
       ELSE
 
-        !FIXME
         !$omp parallel do default(shared) private(i,j,k,a1,a2)
-#ifdef _B4B05R
-        !$acc update host(rho,q3d,ruh,rvh,rmh)
-#else
-        !$acc parallel default(present) private(k) reduction(+:budt) 
+        !$acc parallel default(present) private(k) reduction(+:budt,asq) 
         !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:asq)` here?

src/misclibs.F
```diff
@@ -1490,25 +1413,13 @@ subroutine pdefqtest(rmax,asq,ruh,rvh,rmh,rho,q3d,dum2d1,dum2d2,dum2d3,dum2d4,du
...
-        enddo
-#ifdef _B4B05R
-        !$acc update device(asq)     
-#endif
-
       ENDIF
 
 !----------------------------------------------------------------------
-      !$acc end data
+      !!$acc end data
```
sjsprecious: delete it.

src/misclibs.F
```diff
@@ -4001,47 +3888,24 @@ subroutine calcmass(nstat,rstat,ruh,rvh,rmh,rho)
...
-      !$acc data create(foo)
 
+      tmass=0.0d0 
       !$omp parallel do default(shared) private(i,j,k)
-#ifdef _B4B06R
-      !$acc update host(rho,ruh,rvh,rmh)
-#else
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot)
+      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
       !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:tmass)` here?

src/misclibs.F
```diff
@@ -4093,46 +3957,24 @@ subroutine totmois(nstat,rstat,train,ruh,rvh,rmh,qv,ql,qi,rho)
...
-#ifdef _B4B07R
-      !$acc update host(rho,qv,ql,qi,ruh,rvh,rmh)
-#else
-      !$acc data create(foo)
-
+      tmass=0.0d0
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot)
+      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
       !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:tmass)` here?

src/misclibs.F
```diff
@@ -4157,9 +3999,7 @@ subroutine totmois(nstat,rstat,train,ruh,rvh,rmh,qv,ql,qi,rho)
  
       if(timestats.ge.1) time_stat=time_stat+mytime()
 
-#ifndef _B4B07R
-!$acc end data
-#endif
+!!$acc end data
```
sjsprecious: delete it.

src/misclibs.F
```diff
@@ -4188,45 +4028,25 @@ subroutine totq(nstat,rstat,ruh,rvh,rmh,q,rho,aname)
 
       integer i,j,k
       double precision :: tmass,var
-      double precision, dimension(nk) :: foo
+      !double precision, dimension(nk) :: foo
```
sjsprecious: delete it.

src/misclibs.F
```diff
@@ -4188,45 +4028,25 @@ subroutine totq(nstat,rstat,ruh,rvh,rmh,q,rho,aname)
...
-#ifdef _B4B08R
-      !$acc update host(rho,q,ruh,rvh,rmh)
-#else
-      !$acc data create(foo)
 
+      tmass=0.0d0
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot)
+      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
       !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:tmass)` here?

src/misclibs.F
```diff
@@ -4250,9 +4070,7 @@ subroutine totq(nstat,rstat,ruh,rvh,rmh,q,rho,aname)
 
       if(timestats.ge.1) time_stat=time_stat+mytime()
 
-#ifndef _B4B08R
-!$acc end data
-#endif
+!!$acc end data
```
sjsprecious: delete it.

src/misclibs.F
sherimickelson: Can we delete this line?

src/misclibs.F
```diff
@@ -4285,33 +4103,23 @@ subroutine calcener(nstat,rstat,ruh,rvh,zh,rmh,pi0,th0,rho,ua,va,wa,ppi,tha,   &
...
-#else
-      !$acc parallel default(present) private(i,j,k,u,v,w,tmp,qtot,tem) &
-      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t,ek,ei,ep,le)
-#endif
       !$omp parallel do default(shared)  &
       !$omp private(i,j,k,u,v,w,tmp,qtot,tem)
-#ifndef _B4B
+      !$acc parallel default(present) private(i,j,k,u,v,w,tmp,qtot,tem) &
+      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t,ek,ei,ep,le)
       !$acc loop gang
```
sjsprecious: Do we need to add `reduction(+:ek,ei,ep,le)` here?

src/misclibs.F
```diff
@@ -4428,7 +4223,7 @@ subroutine calcmoe(nstat,rstat,ruh,rvh,rmh,rho,ua,va,wa,qv,ql,qi,vr)
  
       integer i,j,k
       double precision :: tmu,tmv,tmw,qtot,var,tem
-      double precision, dimension(nk) :: foo1,foo2,foo3
+      ! double precision, dimension(nk) :: foo1,foo2,foo3
```
sjsprecious: delete it.

src/misclibs.F
sherimickelson: Can we remove this line?

src/misclibs.F
johnmauff: @sherimickelson.  Yes, we can.  I saw this shortly after I created the PR. I have gone through and since eliminated a number of such lines.

src/misclibs.F
johnmauff: Already done.

src/misclibs.F
```diff
@@ -4554,50 +4339,30 @@ subroutine tmf(nstat,rstat,ruh,rvh,rho,wa)
...
-     !$acc parallel default(present)  private(i,j,k,mf) reduction(+:foo1t,foo2t) &
-     !$acc reduction(+:tmfu,tmfd)
-     !$acc loop gang 
-#endif
+      tmfu=0.0d0
+      tmfd=0.0d0
+      !$omp parallel do default(shared) private(i,j,k,mf)
+      !$acc parallel default(present)  private(i,j,k,mf) reduction(+:foo1t,foo2t) &
+      !$acc reduction(+:tmfu,tmfd)
+      !$acc loop gang
```
sjsprecious, line 4358: Do we need to add `reduction(+:tmfu,tmfd)` here?

src/misclibs.F
```diff
@@ -4625,7 +4390,7 @@ subroutine tmf(nstat,rstat,ruh,rvh,rho,wa)
 #ifdef MPI
       endif
 #endif
-      !$acc end data
+!      !$acc end data
```
sjsprecious: delete it.

sjsprecious (commented):
I have a few similar comments related to the missing `reduction` clause in the `!$acc loop gang` directive. This may or may not address the correctness reported in this PR and I would like to check if it helps.

src/sfcphys.F
johnmauff: done

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Yes we do.  Good catch

src/misclibs.F
johnmauff: Good catch.

src/misclibs.F
johnmauff: Yes, good catch.

src/misclibs.F
johnmauff: Good catch, yes we do.

src/misclibs.F
johnmauff: Yes we do.

src/misclibs.F
johnmauff: Yes, we do.

src/misclibs.F
johnmauff: Yes we do

src/sfcphys.F
johnmauff: Thanks for the catch.

src/misclibs.F
```diff
@@ -4293,25 +4106,19 @@ subroutine calcener(nstat,rstat,ruh,rvh,zh,rmh,pi0,th0,rho,ua,va,wa,ppi,tha,   &
...
-      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t,ek,ei,ep,le)
-#endif
       !$omp parallel do default(shared)  &
       !$omp private(i,j,k,u,v,w,tmp,qtot,tem)
-#ifndef _B4B
-      !$acc loop gang 
-#endif
+      !$acc parallel default(present) private(i,j,k,u,v,w,tmp,qtot,tem) &
+      !$acc reduction(+:foo1t,foo2t,foo3t,foo4t,ek,ei,ep,le)
+      !JMD internal compiler error
```
johnmauff, line 4113: Unfortunately, the addition of a reduction() clause triggered an internal compiler error. So I reverted it to the previous loop structure

src/misclibs.F
johnmauff: Unfortunately, the addition of a reduction() clause generates an internal compiler error.  So I reverted the code to the original loop structure.

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

src/sfcphys.F
johnmauff: Done

src/misclibs.F
```diff
@@ -4583,21 +4370,14 @@ subroutine tmf(nstat,rstat,ruh,rvh,rho,wa)
...
-
       tmfu=0.0d0
       tmfd=0.0d0
-#ifndef _B4B10R
       !$acc loop vector reduction(+:tmfu,tmfd)
-#endif
       do k=1,nk
         tmfu=tmfu+foo1(k)
         tmfd=tmfd+foo2(k)
       enddo
```
sjsprecious: Thanks for reporting the internal compiler error for `tmfu` and `tmfd`. See my previous comment about separating the kernels here.

src/misclibs.F
sjsprecious: Thanks for reporting the internal compiler error. No idea about why it happens. After a second thought, I think it may be safer to write the reductions here as a separate acc parallel region. My main concern is that there is no guarantee for the first k loop to be completed before we execute the second k loop for reduction (I could be wrong though).

sjsprecious (commented):
Thanks for the update. I just have two more suggestions due to the concern of correctness.

johnmauff:
@sjsprecious.  I just updated the description of the PR.  Note that the number of statpack variables that match varies run to run.  I would prefer to keep the new loop structure as eliminates the need to create thread temporary copies for all but the loops that generate the internal compiler error.

sjsprecious:
@johnmauff thanks for the updated verification results. I agree that it is better to keep the new loop structure except the ones with internal compiler error. In addition, I would suggest separating the kernel with internal compiler error into two kernels, as I am concerned about a potential race condition.

johnmauff:
@sjsprecious based on your suggestion, I have broken up the two kernels that generate the internal compiler error.  The statistics are now consistent for two runs.  I have updated the PR description to include the current statistics.

src/misclibs.F
```diff
@@ -4143,14 +4143,14 @@ subroutine calcener(nstat,rstat,ruh,rvh,zh,rmh,pi0,th0,rho,ua,va,wa,ppi,tha,   &
         foo3(k) = foo3t
         foo4(k) = foo4t
       enddo
-      !$acc loop vector reduction(+:ek,ei,ep,le)
+      !$acc end parallel
+      !$acc parallel loop vector default(present) reduction(+:ek,ei,ep,le)
```
sjsprecious: better to add `gang` clause here

src/misclibs.F
```diff
@@ -4143,14 +4143,14 @@ subroutine calcener(nstat,rstat,ruh,rvh,zh,rmh,pi0,th0,rho,ua,va,wa,ppi,tha,   &
...
       do k=1,nk
         ek=ek+foo1(k)
         ei=ei+foo2(k)
         ep=ep+foo3(k)
         le=le+foo4(k)
       enddo
```
sjsprecious, line 4153: better to add the `!$acc end parallel` directive

src/misclibs.F
```diff
@@ -4370,9 +4369,10 @@ subroutine tmf(nstat,rstat,ruh,rvh,rho,wa)
         foo1(k) = foo1t
         foo2(k) = foo2t
       enddo
+      !$acc end parallel
       tmfu=0.0d0
       tmfd=0.0d0
-      !$acc loop vector reduction(+:tmfu,tmfd)
+      !$acc parallel loop vector default(present) reduction(+:tmfu,tmfd)
```
sjsprecious: better to add the `gang` clause here

sjsprecious (commented):
@johnmauff thanks for the update and glad to know that we have consistent results between two runs now, suggesting that there is a race condition in the original implementation. I just have a few more minor comments.

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

src/misclibs.F
johnmauff: Done

sjsprecious (approved):
This PR looks good to me now. Thanks @johnmauff for cleaning up the codes.

sherimickelson (approved):
Thanks @johnmauff .  I think it looks good too.

PR #74, opened 2022-08-17
Add GPU memory usage check interface

sjsprecious:
This PR introduces an interface to check the GPU memory usage for a CM1 run. The Fortran code will call a CUDA API to check the GPU memory usage at the end of simulation.

Note that on Casper, we need to manually link the `cudart` library in the `Makefile` to compile the codes successfully. See the commented-out line for the `FFLAGS_M` option.

`FFLAGS_M = $(OPTS_M) # -L/glade/u/apps/opt/nvhpc/22.2/Linux_x86_64/22.2/cuda/11.6/lib64 -lcudart`

### DIFF
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -1,3 +1,21 @@
+#ifdef _OPENACC
+module cuda_rt
+
+  use iso_c_binding
+
+  interface
+     !
+     integer (c_int) function cudaMemGetInfo(fre, tot) bind(C, name="cudaMemGetInfo")
+       use iso_c_binding
+       implicit none
+       type(c_ptr),value :: fre
+       type(c_ptr),value :: tot
+     end function cudaMemGetInfo
+     !
+  end interface
+
+end module cuda_rt
+#endif
 
       program cm1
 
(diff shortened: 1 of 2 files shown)


src/cm1.F
```diff
@@ -4587,6 +4613,22 @@ program cm1
       print *,'Program terminated normally'
 #endif
 
+#ifdef _OPENACC
```
johnmauff: @sjsprecious instead of always turning this on when _OPENACC is activated, I would instead change the cpp variable so that it is a separate option.  That way if George or David want to run the OPENACC code without using memory profiling everything compiles properly.  Right now if you don't manually change the Makefile the build should fail.

sjsprecious: @johnmauff Thanks for your comments. I have added a new CPP variable named `_CUDAMEM` to enable the GPU memory usage separately. I could confirm that by doing this, I could build a GPU run normally without enabling this functionality.

sherimickelson (approved):

johnmauff (approved):
Thanks, @sjsprecious this now looks good.

PR #75, opened 2022-08-23
New MPI implementation for droplet in CM1

sjsprecious:
This PR will introduce a new MPI implementation for the droplet calculations in CM1.

In the **original** MPI implementation, each MPI rank has to keep a full copy of `pdata` with the length of `nparcels` and perform the droplet calculation on each droplet (even if a droplet does not belong to this MPI rank). An `mpi_allreduce` operation is performed at the end of droplet calculations to ensure that each droplet id updated properly.

In the **new** MPI implementation, each MPI rank only keeps a partial copy of `pdata` with the length of `nparcels/numprocs+buffer_space`, where `buffer_space` is a user-defined space to hold additional droplets in each MPI rank due to the transport. Each MPI rank works on different portion of `pdata` and no `mpi_allreduce` is needed anymore. The `mpi_isend` and `mpi_irecv` interfaces are exploited to handle the droplet movement between different MPI ranks. An `mpi_gather` operation is used to write the `pdata` out to a file.

The verification results for droplet diagnostic (1000 droplets) suggests:
- Original MPI implementation vs. new MPI implementation (both on CPU): consistent 1st-order droplet diagnostic; only `rhoavg` has an average difference larger than 1E-6 for 0th-order droplet diagnostic.
- New MPI implementation, CPU (intel) vs. GPU (nvhpc, 2 MPI ranks + 2 GPUs): consistent 1st-order droplet diagnostic; only `radmax` and `radmin` have an average difference larger than 1E-6 for 0th-order droplet diagnostic.

There is a correctness issue when we increase the droplet number to ~1M. The droplet diagnostic does not agree between the original and new MPI implementations either on CPU or GPU. The difference also exists between CPU and GPU when only the original MPI implementation is used. The fix is still working in progress.

### DIFF
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -2230,13 +2230,17 @@
       allocate( reqt(d3t) )
       reqt = 0
 
+#ifdef MPI
       nparcels_mpi = ceiling(nparcels*1.0/numprocs+pdata_buffer*nparcels)
+#else
+      nparcels_mpi = nparcels
+#endif
 
       !$acc update device (nparcels_mpi)
 
       allocate(  pdata(nparcels_mpi,npvals) )
       pdata = 0.0
-      allocate(   ploc(nparcels,  3   ) )
+      allocate(   ploc(nparcels_mpi,  3   ) )
       ploc = 0.0
       allocate(  pdata_locind(nparcels_mpi,3) )
       pdata_locind = undefined_index
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -3488,7 +3488,18 @@
 
 !#endif
         !Also compute the droplet diagnostics at the same frequency as statpack
-        if(myid == 0) call droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind)
+
+        ! JS: with the new MPI implementation, we can:
+        !     1) use MPI_GATHER to collect the droplet information from different
+        !        MPI ranks first and only the root rank calculates the diagnostic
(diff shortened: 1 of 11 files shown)


johnmauff (approved):
@sjsprecious.  I think this PR generally looks good I do have a couple of recommendations:

  (1) You might want to note in the message that correctness is only achieved for configurations where you use a small number of droplets.  There are correctness issues for a larger number of droplets (>1M) is still a work in progress.

   (2) Use of the variable nparcels_mpi is a little confusing especially when the code is compiled without MPI support.  Maybe it needs to be something more generic like nparcelsLocal  This is just a suggestion that can be applied to subsequent revisions of the code.

sjsprecious:
> @sjsprecious. I think this PR generally looks good I do have a couple of recommendations:
> 
> (1) You might want to note in the message that correctness is only achieved for configurations where you use a small number of droplets. There are correctness issues for a larger number of droplets (>1M) is still a work in progress.
> 
> (2) Use of the variable nparcels_mpi is a little confusing especially when the code is compiled without MPI support. Maybe it needs to be something more generic like nparcelsLocal This is just a suggestion that can be applied to subsequent revisions of the code.

Thanks @johnmauff . I have updated the variable name from `nparcels` to `nparcelsLocal`. I have also updated the correctness issue for 1M droplets in the message section.

sherimickelson (approved):
Thanks for your great work on this @sjsprecious

sjsprecious:
@sherimickelson @johnmauff I have pushed John's bug fix for `param.F` as well.

PR #76, opened 2022-08-29
Eliminate stop

johnmauff:
Currently, the gpu-opt has a number of 'stop' calls that protect code sections that have not been ported to GPU.  This PR addresses a number of these in the following manner:
    (1)  Port several sections of code to OPENACC
    (2)  Reduce the level of severity from a fatal error to a warning to indicate code that has not been formally verified
    (3)  Deleted a number of #ifdef OPENACC that were no longer needed.
This PR also addresses a couple of bugs including:
     (A)The length of the padding for droplets that move between MPI is too short for a small number of particles
     (B) Compiler error zf_tmp when OPENACC is enabled and MPI is disabled.

This PR has the same validation characteristics as the current gpu-opt branch

### DIFF
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -2701,9 +2701,6 @@ end module cuda_rt
           ndt  = 0
           adt  = 0.0
           acfl = 0.0
-#ifdef _OPENACC
-          !print *, 'WARNING: getnewdt not supported on OpenACC' 
-#endif
           call   getnewdt(ndt,dt,dtlast,adt,acfl,dbldt,                                 &
                           mtime,stattim,taptim,rsttim,prcltim,diagtim,azimavgtim,       &
                           dorestart,dowriteout,dostat,doprclout,dotdwrite,doazimwrite,  &
@@ -3434,7 +3431,7 @@ end module cuda_rt
       IF( (idiff.ge.1).and.(difforder.eq.2) )THEN
         !  get stress terms for explicit diffusion scheme:
 #ifdef _OPENACC
-        stop 'ERROR: diff2def not supported on OpenACC version' 
+        stop 'ERROR: diff2def not supported on OPENACC version' 
 #endif
         call diff2def(uh,arh1,arh2,uf,arf1,arf2,vh,vf,mh,c1,c2,mf,ust,zntmp,u1,v1,s1,  &
                       divx,rho,rr,rf,t11,t12,t13,t22,t23,t33,u3d,v3d,w3d,dissten)
@@ -3458,9 +3455,6 @@ end module cuda_rt
 
         IF( adapt_dt.eq.1 .and. dosolve )THEN
 
-#ifdef _OPENACC
-          !print *,'WARNING: getnewdt not supported on OpenACC' 
-#endif
           call   getnewdt(ndt,dt,dtlast,adt,acfl,dbldt,                                 &
                           mtime,stattim,taptim,rsttim,prcltim,diagtim,azimavgtim,       &
                           dorestart,dowriteout,dostat,doprclout,dotdwrite,doazimwrite,  &
@@ -3854,7 +3848,7 @@ end module cuda_rt
       if( doazimwrite )then
 
 #ifdef _OPENACC
-        stop 'ERROR: azimavg not supported in OpenACC'
+        stop 'ERROR: azimavg not supported in OPENACC'
 #endif
         call   azimavg(nstep,mtime,nwritea,arecs,arecw,qname,dt,dosfcflx,      &
                    icrs,icenter,jcenter,xcenter,ycenter,                       &
@@ -3908,7 +3902,7 @@ end module cuda_rt
       if( dohifrqwrite )then
 
 #ifdef _OPENACC
-        stop 'ERROR: writeout_hifrq not supported in OpenACC'
+        stop 'ERROR: writeout_hifrq not supported in OPENACC'
 #endif
         call   writeout_hifrq(                                                 &
                    nstep,mtime,nwriteh,qname,dt,dosfcflx,                      &
@@ -3975,7 +3969,7 @@ end module cuda_rt
           enddo
         endif
 #ifdef _OPENACC
-        print *,'ERROR: write_restart not supported in OpenACC'
+        print *,'ERROR: write_restart not supported in OPENACC'
 #endif
         call     write_restart(nstep,srec,sirec,urec,vrec,wrec,nrec,mrec,prec,      &
                                trecs,trecw,arecs,arecw,                             &
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -284,6 +284,7 @@ end module cuda_rt
       real :: mp_total,minval,temx,temy,temni,temnj
       double precision :: tstart,tend
       integer, dimension(:), allocatable :: isum,jsum
+      integer :: padding
 #endif
 #ifdef _OPENACC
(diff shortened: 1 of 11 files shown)


sherimickelson (approved):

src/cm1.F
```diff
@@ -2259,7 +2260,8 @@ program cm1
       reqt = 0
 
 #ifdef MPI
-      nparcelsLocal = ceiling(nparcels*1.0/numprocs+pdata_buffer*nparcels)
+      padding = max(pdata_buffer*nparcels,100)
+      nparcelsLocal = ceiling((nparcels*1.0/numprocs)+padding)
```
sjsprecious, line 2264: I like this design. Save my time to change `pdata_buffer` for small droplet number.

src/interp_routines.F
```diff
@@ -440,14 +440,10 @@ subroutine vinterp_weno5(stag,ix,jy,kz,c1,c2,rrw,dumz,a,pdef,pdefweno,weps)
...
 
   ELSEIF( stag.eq.4 )THEN
-#ifdef _OPENACC
-    stop 'vinterp_weno5:  stag == 4 not supported on OpenACC'
-#endif
 
-!$omp parallel do default(shared)   &
-!$omp private(i,j,k,wbar,cc1,cc2)
+    !$omp parallel do default(shared)  private(i,j,k,wbar)
+    !$acc parallel loop gang vector collapse(3) default(present)
```
sjsprecious, line 445: According to the PR description, adding lots of OpenACC directives below does not change the verification answers. Does it mean that these calculations are not called at all?

src/interp_routines.F
```diff
@@ -554,10 +562,13 @@ subroutine vinterp_flx5(stag,ix,jy,kz,c1,c2,rrw,dumz,a)
...
       enddo
       enddo
+    ENDDO
 
-      !----
+    !----
 
-      k = 3
+    k = 3
+    !$acc parallel loop gang vector collapse(2) default(present)
```
sjsprecious, line 571: Do we need OpenMP directive here?

src/interp_routines.F
```diff
@@ -566,8 +577,11 @@ subroutine vinterp_flx5(stag,ix,jy,kz,c1,c2,rrw,dumz,a)
           dumz(i,j,k) = flx5(a(i,j,k+2),a(i,j,k+1),a(i,j,k  ),a(i,j,k-1),a(i,j,k-2))
         endif
       enddo
+    ENDDO
 
-      k = nk-1
+    k = nk-1
+    !$acc parallel loop gang vector collapse(2) default(present)
```
sjsprecious, line 585: Do we need OpenMP directive here?

src/interp_routines.F
```diff
@@ -576,10 +590,13 @@ subroutine vinterp_flx5(stag,ix,jy,kz,c1,c2,rrw,dumz,a)
           dumz(i,j,k) = flx3(a(i,j,k+1),a(i,j,k  ),a(i,j,k-1))
         endif
       enddo
+    ENDDO
 
       !----
 
-      k = 2
+    k = 2
+    !$acc parallel loop gang vector collapse(2) default(present)
```
sjsprecious, line 601: Do we need OpenMP directive here?

src/interp_routines.F
```diff
@@ -588,8 +605,11 @@ subroutine vinterp_flx5(stag,ix,jy,kz,c1,c2,rrw,dumz,a)
           dumz(i,j,k) = flx3(a(i,j,k+1),a(i,j,k  ),a(i,j,k-1))
         endif
       enddo
+    ENDDO
 
-      k = nk
+    k = nk
+    !$acc parallel loop gang vector collapse(2) default(present)
```
sjsprecious, line 615: Do we need OpenMP directive here?

src/misclibs.F
```diff
@@ -393,8 +393,11 @@ subroutine convinitu(myid,ib,ie,jb,je,kb,ke,ni,nj,nk,ibw,ibe,   &
       !$acc declare present(xf,yh,zh,u0,u3d)
 
 !!!      if(myid.eq.0) print *,'    convinitu '
+#ifdef _OPENACC
+      print *,'WARNING: OPENACC version of convinitu never verified'
+#endif
       !$omp parallel do default(shared) private(i,j,k,term1,term2,term3,term4,umo)
-      !$acc parallel loop gang vector default(present) private(i,j,k,term1,term2,term3,term4,umo)
+      !$acc parallel loop gang vector collapse(3) default(present)
```
sjsprecious, line 399: I think we have race condition here for both OpenMP and OpenACC here?

src/misclibs.F
```diff
@@ -434,9 +437,12 @@ subroutine convinitv(myid,ib,ie,jb,je,kb,ke,ni,nj,nk,ibs,ibn,   &
       real :: term1,term2,term3,term4,vmo
       !$acc declare present(xh,yf,zh,v0,v3d)
 
+#ifdef _OPENACC
+      print *,'WARNING: OPENACC version of convinitv never verified'
+#endif
 !!!      if(myid.eq.0) print *,'    convinitv '
       !$omp parallel do default(shared) private(i,j,k,term1,term2,term3,term4,vmo)
-      !$acc parallel loop gang vector default(present) private(i,j,k,term1,term2,term3,term4,vmo)
+      !$acc parallel loop gang vector collapse(3) default(present)
```
sjsprecious, line 443: I think we have race condition here for both OpenMP and OpenACC here?

sjsprecious (changes requested):
Though the verification test results are not changed, I still have a few comments about this PR, especially for some potential race conditions (correct me if I am wrong).

src/interp_routines.F
johnmauff: @sjsprecious, none of the modified code in this PR has ever been executed using any of the existing namelists that we have used.

src/interp_routines.F
johnmauff: Yes we do, I have added in OpenMP directives.

src/interp_routines.F
johnmauff: Yes we do, I have added in OpenMP directives.

src/interp_routines.F
johnmauff: Yes we do, I have added in OpenMP directives

src/interp_routines.F
johnmauff: Yes we do, I have added in OpenMP directives.

src/misclibs.F
johnmauff: @sjsprecious, I am not seeing the race condition in this loop.  Could you add more details to your comment?

src/misclibs.F
johnmauff: @sjsprecious, I am not seeing the race condition in this loop.  Could you add more details to your comment?

src/misclibs.F
sjsprecious: My bad! I thought `term1`, `term2`, `term3`, `term4` and `umo` were returned variables. You are right that there is no race condition here. A good example of not reviewing PR during the midnight.

src/misclibs.F
sjsprecious: You are right that there is no race condition here.

sjsprecious (approved):
My comments are wrong and the PR looks good to me now.

PR #77, opened 2022-09-01
Fix nvhpc build

johnmauff:
This PR appears to address two issues, the inability to build CM1 on Gust with the NVHPC 22.7 compiler, and the slow leaking of droplets for very large droplet counts.  Note that for the standard verification namelist 'namelist.baseline.dx=40m.short' where nparcels=1000 the verification statistics are unchanged. When nparcels=1000000, the total number of droplets is unchanged for the entire run.  The changes are as follows:

    (1) Properly set several module variables on the device (dx,dy,mywest,myeast,mysouth,mynorth,myse,mysw,myne,mynw). While a compiler warning message was generated, it was missed due to the verbose output generated by -Minfo=accel compiler flag. Because dx and dy were not set properly, droplets near the boundary between different MPI ranks were being lost.

    (2) Changed the makefile to support both A100 and V100 architectures.  The -Minfo=accel compiler flag was removed to allow warning messages to be easily identified.

   (3)  Use of the array constructor appears to generate an internal compiler error under certain circumstances.  For example: fv2 = (/1.,1./) was generated the internal compiler error.  Replaced use of the array constructor with explicit assignment.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -36,6 +36,7 @@ USE_OPENACC_L = $(shell echo $(USE_OPENACC) | tr A-Z a-z)
 USE_NETCDF_L  = $(shell echo $(USE_NETCDF) | tr A-Z a-z)
 DEBUG_L       = $(shell echo $(DEBUG) | tr A-Z a-z)
 USE_CUDAMEM_L = $(shell echo $(USE_CUDAMEM) | tr A-Z a-z)
+GPU_TYPE_L    = $(shell echo $(GPU_TYPE) | tr A-Z a-z)
 
 #-----------------------------------------------------------------------------
 #  initialize some options as empty
@@ -99,14 +100,24 @@ ifeq (${FC},nvfortran)
     endif
     ifeq (${USE_OPENACC_L},true)
         ifeq (${DEBUG_L},true)
-            OPTS += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
-            OPTS_M += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
+            ifeq (${GPU_TYPE_L},a100)
+                OPTS += -acc -gpu=cc80,lineinfo,nofma,autocompare,math_uniform
+                OPTS_M += -acc -gpu=cc80,lineinfo,nofma,autocompare,math_uniform
+            else
+                OPTS += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
+                OPTS_M += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
+            endif
         else
-	    # Potentially interesting non-default compiler flags:: 
-	    #     -Minfo=accel    Provides verbose output for accelerator compilation phase
-	    #     -Mpcast         Enables the use of PCAST functionality
-            OPTS += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform
-            OPTS_M += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform,managed
+            # Potentially interesting non-default compiler flags:: 
+            #     -Minfo=accel    Provides verbose output for accelerator compilation phase
+            #     -Mpcast         Enables the use of PCAST functionality
+            ifeq (${GPU_TYPE_L},a100)
+                OPTS += -acc -gpu=cc80,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
+                OPTS_M += -acc -gpu=cc80,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
+            else
+                OPTS += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
+                OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
+            endif
         endif
         DM       += -D_OPENACC
     endif
@@ -153,7 +164,9 @@ endif
 #  Intel compiler
 #-----------------------------------------------------------------------------
 ifeq (${FC},ifort)
-    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    # replace -xHost with -march=core-avx2 to make sure the same code builds on
+    # both Intel and AMD CPUs 
+    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
     CPP          += cpp -C -P -traditional -Wno-invalid-pp-token -ffreestanding
     ifeq (${USE_OPENMP_L},true)
         OPTS     += -qopenmp
--- a/src/Makefile
+++ b/src/Makefile
@@ -102,8 +102,11 @@ ifeq (${FC},nvfortran)
             OPTS += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
             OPTS_M += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
         else
-            OPTS += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
-            OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
+	    # Potentially interesting non-default compiler flags:: 
+	    #     -Minfo=accel    Provides verbose output for accelerator compilation phase
+	    #     -Mpcast         Enables the use of PCAST functionality
+            OPTS += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform
+            OPTS_M += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform,managed
         endif
         DM       += -D_OPENACC
     endif
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -141,7 +141,7 @@
       !$acc      copyin (randomNumbers,randomNumbers%state, &
       !$acc              pdata_neighbor)
 #else
-      !$acc data create (ta,zf_tmp) &
+      !$acc data create (ta) &
       !$acc      copyin (randomNumbers,randomNumbers%state)
 #endif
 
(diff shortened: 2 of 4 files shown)


sherimickelson (approved):

src/param.F
```diff
@@ -228,8 +228,8 @@ subroutine param(namelist_input,                                  &
 #endif
 
       call bcast_log(terrain_flag,device=.true.)
-      call bcast_real(dx)
-      call bcast_real(dy)
+      call bcast_real(dx,device=.true.)
+      call bcast_real(dy,device=.true.)
```
sjsprecious, line 232: My bad. I did not realize that these two variables are now used in an OpenACC subroutine and should be updated manually. Thanks for catching this error!

src/param.F
```diff
@@ -6284,6 +6284,7 @@ subroutine param(namelist_input,                                  &
       mysouth = nabor(myi,   myj-1, nodex, nodey)
       myeast  = nabor(myi+1, myj,   nodex, nodey)
       mywest  = nabor(myi-1, myj,   nodex, nodey)
+      !$acc update device(mynorth,mysouth,myeast,mywest)
```
sjsprecious, line 6287: Thanks for catching this error!

src/param.F
```diff
@@ -6295,6 +6296,7 @@ subroutine param(namelist_input,                                  &
       mynw = nabor(myi-1, myj+1,   nodex, nodey)
       myne = nabor(myi+1, myj+1,   nodex, nodey)
       myse = nabor(myi+1, myj-1,   nodex, nodey)
+      !$acc update device(mysw,mynw,myne,myse)
```
sjsprecious, line 6299: Thanks for catching this error!

src/droplet.F
```diff
@@ -2201,7 +2201,11 @@ subroutine gauss_newton_2d(diffnorm,dt_nondim,taup_scale,rt_start,rt_zeros,flag,
 
         v1(1) = rt_start(1)
         v1(2) = rt_start(2)
-        fv2 = (/1., 1./)
+         
+        fv2(:) = 1.
+        !JMD-FIXME:  The following line appears to generate the following compiler error:
+        !JMD-FIXME     Internal compiler error. Basic LLVM base data type required 
+        !fv2 = (/1., 1./)
```
sjsprecious, line 2209: I thought that I had rewritten all these operations explicitly when I ported BE solver to GPU and encountered a similar issue, but clearly I did not. Thanks for catching this error!

src/droplet.F
```diff
@@ -2201,7 +2201,11 @@ subroutine gauss_newton_2d(diffnorm,dt_nondim,taup_scale,rt_start,rt_zeros,flag,
 
         v1(1) = rt_start(1)
         v1(2) = rt_start(2)
-        fv2 = (/1., 1./)
+         
+        fv2(:) = 1.
```
sjsprecious: Will it be better to write `fv2(1) = 1.` and  `fv2(2) = 1.`?

src/droplet.F
```diff
@@ -2282,12 +2286,20 @@ subroutine LV_solver(diffnorm,dt_nondim,taup_scale,rt_start,rt_zeros,flag,prsval
...
+        I(1,2)=0.0
+        I(2,2)=1.0
         iterations = 0
         flag = 0
         v1(1) = rt_start(1)
         v1(2) = rt_start(2)
-        fv2 = (/1., 1./)
+        !JMD-FIXME generates an internal compiler error
+        ! fv2 = (/1., 1./)
+        fv2(:) = 1.0
```
sjsprecious: Will it be better to write `fv2(1) = 1.0` and  `fv2(2) = 1.0`?

src/Makefile
```diff
@@ -102,8 +102,11 @@ ifeq (${FC},nvfortran)
             OPTS += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
             OPTS_M += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
         else
-            OPTS += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
-            OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
+	    # Potentially interesting non-default compiler flags:: 
+	    #     -Minfo=accel    Provides verbose output for accelerator compilation phase
+	    #     -Mpcast         Enables the use of PCAST functionality
+            OPTS += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform
+            OPTS_M += -acc -gpu=cc70,cc80,lineinfo,nofma,math_uniform,managed
```
sjsprecious: Based on my understanding, doing these will make the executable file much larger in size. I have introduced a new compiler option `GPU_TYPE` in the Makefile and it looks like:

```
GPU_TYPE_L    = $(shell echo $(GPU_TYPE) | tr A-Z a-z)

...

ifeq (${USE_OPENACC_L},true)
    ifeq (${DEBUG_L},true)
        ifeq (${GPU_TYPE_L},a100)
            OPTS += -acc -gpu=cc80,lineinfo,nofma,autocompare,math_uniform
            OPTS_M += -acc -gpu=cc80,lineinfo,nofma,autocompare,math_uniform
        else
            OPTS += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
            OPTS_M += -acc -gpu=cc70,lineinfo,nofma,autocompare,math_uniform
        endif
    else
        ifeq (${GPU_TYPE_L},a100)
            OPTS += -acc -gpu=cc80,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
            OPTS_M += -acc -gpu=cc80,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
        else
            OPTS += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
            OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
        endif
    endif
    DM       += -D_OPENACC
endif
```

Will this implementation be better?

sjsprecious (commented):
Thanks for catching those GPU errors and apologize for my sloppy implementation. Do I understand it correctly that with your changes, we do not need to comment out the "conflict check" section and still conserve the total droplets for the 1 million case? Does it reduce the inconsistency of droplet diagnostic for the 1 million case?

src/param.F
johnmauff: @ no problem.  I was a bit surprised why the compiler did not flag it as a fatal error.

sjsprecious: Probably because `dx` and `dy` are used in both acc routine and acc kernel?

sjsprecious:
And shall we update the check of droplets that will jump over the MPI ranks in the BE solver (https://github.com/george-bryan/CM1/blob/fix_nvhpc_build/src/droplet.F#L1240) based on George's suggestion? An example will be like:

```
      if ( (x3d .ge. xf(1)) .and. (x3d .lt. xf(ni+1)) .and.  &
           (y3d .ge. yf(1)) .and. (y3d .lt. yf(nj+1)) ) then
           neighbor = undefined_index   ! undefined_index means this droplet
                                                          ! stays within the same MPI region
      else if ( (x3d .lt. xf(1)) .and. (y3d .ge. yf(1)) .and. (y3d .lt. yf(nj+1)) ) then
           if ( x3d .lt. xf(1)-dx*ni ) then
              stop "Jump over the west nearest neighbor"
           end if
           neighbor = mywest
      else if ( (x3d .lt. xf(1)) .and. (y3d .lt. yf(1)) ) then
           if ( (x3d .lt. xf(1)-dx*ni) .or. (y3d .lt. yf(1)-dy*nj) ) then
              stop "Jump over the southwest nearest neighbor"
           end if
           neighbor = mysw
      else if ( (x3d .lt. xf(1)) .and. (y3d .ge. yf(nj+1)) ) then
           if ( (x3d .lt. xf(1)-dx*ni) .or. (y3d .ge. yf(nj+1)+dy*nj) ) then
              stop "Jump over the northwest nearest neighbor"
           end if
           neighbor = mynw
      else if ( (x3d .ge. xf(ni+1)) .and. (y3d .ge. yf(1)) .and. (y3d .lt. yf(nj+1)) ) then
           if ( x3d .gt. xf(ni+1)+dx*ni ) then
              stop "Jump over the east nearest neighbor"
           end if
           neighbor = myeast
      else if ( (x3d .ge. xf(ni+1)) .and. (y3d .lt. yf(1)) ) then
           if ( (x3d .gt. xf(ni+1)+dx*ni) .or. (y3d .lt. yf(1)-dy*nj) ) then
              stop "Jump over the southeast nearest neighbor"
           end if
           neighbor = myse
      else if ( (x3d .ge. xf(ni+1)) .and. (y3d .ge. yf(nj+1)) ) then
           if ( (x3d .gt. xf(ni+1)+dx*ni) .or. (y3d .gt. yf(nj+1)+dy*nj) ) then
              stop "Jump over the northeast nearest neighbor"
           end if
           neighbor = myne
      else if ( (x3d .ge. xf(1)) .and. (x3d .lt. xf(ni+1)) .and. (y3d .lt. yf(1)) ) then
           if ( y3d .lt. yf(1)-dy*nj ) then
              stop "Jump over the south nearest neighbor"
           end if
           neighbor = mysouth
      else if ( (x3d .ge. xf(1)) .and. (x3d .lt. xf(ni+1)) .and. (y3d .ge. yf(nj+1)) ) then
           if ( y3d .gt. yf(nj+1)+dy*nj ) then
              stop "Jump over the north nearest neighbor"
           end if
           neighbor = mynorth
      else
           stop "This section should be never entered!"
      end if
```

src/Makefile
johnmauff: @sjsprecious if you want to go ahead and modify the Makefile either for this PR or for a future one I would be fine with these changes.

src/droplet.F
johnmauff: Just updated the code to eliminate the array syntax.

src/droplet.F
johnmauff: Just updated the code to eliminate the array syntax

johnmauff:
@sjsprecious I would recommend that the changes to the droplet code based on George's suggestion be a separate PR.

src/Makefile
sjsprecious: @johnmauff Thanks John. I have just pushed the changes of Makefile to your branch.

sjsprecious:
> @sjsprecious I would recommend that the changes to the droplet code based on George's suggestion be a separate PR.

Sounds reasonable.

sjsprecious:
@johnmauff Thanks for your bug fix. I could confirm that I am able to build an MPI+GPU case on Gust with your PR. There is still a droplet leak for the 1 million case even with your changes, as long as I uncomment the "conflict" check.

sjsprecious (approved):

PR #78, opened 2022-09-09
Fix the race condition in droplet for CM1 GPU code

sjsprecious:
This PR fixes the race condition when calculating **qten** and **thten** in the GPU version of **droplet_driver** subroutine.

Thanks @johnmauff for catching this bug.

It will blow up the GPU memory if we simply add **qten** and **thten** to an OpenACC reduction clause and thus I fix it in the current way. I am open to other options that could resolve this issue smarter.

Verification results:

1. CPU (intel, original) vs. CPU (intel, revised)
- Identical metric and droplet diagnostic output, which suggests that these changes do not affect the CPU results as expected.

2. CPU (intel, original) vs. GPU (nvhpc, 2 MPI + 2 V100 GPU, managed memory, revised)
- Identical variables: 42
- variables with red_diff > 1.e-6: 25
- variables with red_diff <= 1.e-6: 16
- four variables with largest error: KSVMAX, KSHMAX, VOR1KM, VOR2KM

The metric output from the GPU run now agrees much better with that from a CPU run, compared to the result from the current **gpu-opt** branch (`c79cd6e`).

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -96,8 +96,10 @@
       integer :: nrkp
       real :: dt2,uu1,vv1,ww1
       real :: sig3d,th_tmp,rand
-      real :: dvpdt1,dvpdt2,dvpdt3,drpdt,dtpdt,dmvdt
-      real :: Nup,rhop0,taup0,rp0,salinity,part_grav1,part_grav2,part_grav3,rhoval,tval
+      real, dimension(nparcelsLocal) :: dvpdt1,dvpdt2,dvpdt3,drpdt,dtpdt,dmvdt, &
+                                        rhoval,rp0,taup0,rhop0,tval, &
+                                        iflag_tmp,jflag_tmp,kflag_tmp
+      real :: Nup,salinity,part_grav1,part_grav2,part_grav3
       real :: xv,yv,zv,dV,wtx,wty,wtz,wtt
       real :: esl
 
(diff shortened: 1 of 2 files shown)


sherimickelson (approved):
Thanks, @sjsprecious for fixing this!

src/droplet.F
```diff
@@ -408,15 +416,15 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
 
-      call rk2_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,ruh,yf,yh,vh,rvh,zf,zh,zs,znt,sigma,sigmaf,ua,va,wa,qa,ta,rho,prs,x3d,y3d,z3d,sig3d,rhoval,tval,Nup,rhop0,taup0,rp0,part_grav1,part_grav2,part_grav3,debug)
+      call rk2_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,ruh,yf,yh,vh,rvh,zf,zh,zs,znt,sigma,sigmaf,ua,va,wa,qa,ta,rho,prs,x3d,y3d,z3d,sig3d,rhoval(np),tval(np),Nup,rhop0(np),taup0(np),rp0(np),part_grav1,part_grav2,part_grav3,debug)
 
       ELSEIF (idropmethod.eq.1) THEN  !Implicit backward Euler for the droplet integration
 
 #ifdef MPI
-      call BE_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,ruh,yf,yh,vh,rvh,zf,zh,zs,znt,sigma,sigmaf,ua,va,wa,qa,ta,rho,prs,x3d,y3d,z3d,sig3d,rhoval,tval,Nup,rhop0,taup0,rp0,part_grav1,part_grav2,part_grav3,debug,num100,num1000,neighbor)
+      call BE_integration(dt,np,pdata,pdata_locind,iflag,jflag,kflag,xf,xh,uh,ruh,yf,yh,vh,rvh,zf,zh,zs,znt,sigma,sigmaf,ua,va,wa,qa,ta,rho,prs,x3d,y3d,z3d,sig3d,rhoval(np),tval(np),Nup,rhop0(np),taup0(np),rp0(np),part_grav1,part_grav2,part_grav3,debug,num100,num1000,neighbor)
```
johnmauff, line 424: @sjsprecious Personally I think it is awkward to have the interface to BE_integration be different depending on if the code is MPI or not.  I would prefer that this detail is not present in the interface.  I don't think it is necessary to change this interface for this PR, but I am hoping that future code versions could have a single interface.

johnmauff (approved):
@sjsprecious Thanks for coming up with a quick fix for this bug.  I am glad to hear that it has had an impactful change on code correctness.

src/droplet.F
sjsprecious: Thanks @johnmauff. A quick fix will be passing `neighbor` variable anyway and it won't be used in a serial code. And I just need to wrap the `pdata_neighbor(np) = neighbor` line with the `#ifdef MPI` preprocessor variable.

johnmauff: Thanks, @sjsprecious.  I think that would make it easier to read.

PR #79, opened 2022-09-15
Fix tmass and tmf calculations

johnmauff:
This PR request reactivates the calculation of the TMFU, TMFD, and TMASS values for the statpack.F90.  The correctness tests are different as it now adds three new fields.  The statistics are as follows.  Note that while the TMFU and TMFD are outside the relative difference of >1e-06 they only slightly fail this threshold tolerance.  Based on the simplicity of this code which is fundamentally a reduction, we view these differences as acceptable.

44 stat variables are identical.
27 stat variables show a mean relative difference > 1e-06
15 stat variables show a mean relative difference <= 1e-06

### DIFF
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -3899,13 +3899,10 @@
 
       tmass=0.0d0 
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel loop default(present) collapse(3) reduction(+:tmass)
+      !$acc parallel loop gang vector default(present) collapse(3) reduction(+:tmass)
       do k=1,nk
-        !foot=0.0d0
-        !!$acc loop vector collapse(2) reduction(+:foot)
         do j=1,nj
         do i=1,ni
-          ! foot=foot+rho(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
           tmass=tmass+rho(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
         enddo
         enddo
--- a/src/misclibs.F
+++ b/src/misclibs.F
@@ -3899,17 +3899,16 @@
 
       tmass=0.0d0 
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
-      !$acc loop gang reduction(+:tmass)
+      !$acc parallel loop default(present) collapse(3) reduction(+:tmass)
       do k=1,nk
-        foot=0.0d0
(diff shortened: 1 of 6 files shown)


src/misclibs.F
```diff
@@ -3899,17 +3899,16 @@ subroutine calcmass(nstat,rstat,ruh,rvh,rmh,rho)
...
       tmass=0.0d0 
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
-      !$acc loop gang reduction(+:tmass)
+      !$acc parallel loop default(present) collapse(3) reduction(+:tmass)
       do k=1,nk
-        foot=0.0d0
-        !$acc loop vector collapse(2) reduction(+:foot)
+        !foot=0.0d0
+        !!$acc loop vector collapse(2) reduction(+:foot)
```
sjsprecious: remove these two lines

src/misclibs.F
```diff
@@ -3899,17 +3899,16 @@ subroutine calcmass(nstat,rstat,ruh,rvh,rmh,rho)
...
         do j=1,nj
         do i=1,ni
-          foot=foot+rho(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
+          ! foot=foot+rho(i,j,k)*ruh(i)*rvh(j)*rmh(i,j,k)
```
sjsprecious: remove this line

src/misclibs.F
```diff
@@ -3899,17 +3899,16 @@ subroutine calcmass(nstat,rstat,ruh,rvh,rmh,rho)
 
       tmass=0.0d0 
       !$omp parallel do default(shared) private(i,j,k)
-      !$acc parallel default(present) private(i,j,k) reduction(+:foot,tmass)
-      !$acc loop gang reduction(+:tmass)
+      !$acc parallel loop default(present) collapse(3) reduction(+:tmass)
```
sjsprecious: Thanks for catching this and it looks fine to collapse the three nested loops. I would suggest adding `gang vector` clause after the `loop` directive for consistency.

sjsprecious (approved):
Thanks for simplifying some GPU codes. The PR looks good to me and I just have a few minor comments to delete those unused codes.

src/misclibs.F
johnmauff: Thanks for the catch.

src/misclibs.F
johnmauff: @sjsprecious Thanks for catching these several cleanup items.

sherimickelson (approved):

PR #80, opened 2022-09-16
David physics 2

drichte2:
I have re-done the two-way coupling, and now it is all done in a new routine "project_feedback". Unfortunately I didn't realize that Jian and John were working on fixing the race condition at the same time, so I had to undo some of the things that were recently changed in order to get this to compile. The call to "project_feedback" looks like it'll have to be moved to outside of the np loop, but I wasn't sure if I could do that properly and get the acc commands right. 

Also note that this compiles, but does not run. It crashes almost immediately, and in a section of code that doesn't involve the droplets.


sjsprecious:
@johnmauff @sherimickelson I have merged the latest `gpu-opt` branch into David's branch and made the following changes:
- Add the missing variables to the OpenACC directives in `droplet.F`
- Remove the unused RK2 solver and other codes from `droplet.F`
- Revert the value of `rstfrq` namelist variable from **40** back to **10800** in the `namelist.baseline.dx=40m.short
` namelist. Otherwise the simulation will crash. 

With all these changes above, I perform the verification tests using the `namelist.baseline.dx=40m.short` namelist. The result indicates:
- The number of `stat` output is different before merging `gpu-opt` branch (commit: `bfaca49`) and after merging `gpu-opt` branch (commit: `fd091c0`), as somehow David's branch did not include John's changes at #79 . Based on my visual comparison, the existing `stat` output is identical before and after merging `gpu-opt` branch.
- The 0th- and 1st-order droplet diagnostics agree well before and after merging `gpu-opt` branch.

sherimickelson (approved):

johnmauff:
@sjsprecious I don't quite understand the issue with PR #79.  Are these changes missing in this PR?

sjsprecious:
> @sjsprecious I don't quite understand the issue with PR #79. Are these changes missing in this PR?

@johnmauff The changes in the PR #79 are not included when David originally issued this PR. They are in this PR now after I merge the `gpu-opt` branch. But the number of "stat" output is then different between David's original PR (serve as the trusted baseline) and this up-to-date PR. Hope this clarifies.

johnmauff:
@sjsprecious Thank you for the clarification, that helps.  I am now fine with this PR.

johnmauff (approved):
@sjsprecious Looks good to me.

sjsprecious (approved):

PR #81, opened 2022-09-19
Enable write_restart

johnmauff:
This PR adds a number of update host() directives to the write_restart() subroutine.  Currently, this code both compiles and runs.  A future PR will add the necessary namelist to verify correctness.


sherimickelson (approved):

sjsprecious:
@johnmauff I commented out the I/O for the droplet properties (see details at https://github.com/george-bryan/CM1/blob/gpu-opt/src/parcel.F#L1799) since the new MPI interface will include some garbage information for each MPI rank. Does it affect the proper function of the restart run?

PR #83, opened 2022-09-20
Compact the pdata to make active droplets contiguous

sjsprecious:
This PR compacts the **pdata** to make sure that the active droplets for each MPI rank is contiguous after calling the **droplet_driver** subroutine. Therefore, there will be no *holes* in the **pdata** after the MPI communication.

With these changes, we could perform the droplet calculation for the active droplets only rather than the whole **nparcelsLocal** slots of **pdata**.

The verification tests show a consistent result compared to those using the latest **gpu-opt** branch (commit: `ff1374`)


sherimickelson (approved):

src/droplet.F
```diff
@@ -712,7 +556,45 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+
+    ! there will be remaining "holes" in the "pdata" only when more
+    ! droplets leave rather than enter the current MPI region
+    if ( sum(num2) .lt. sum(num1) ) then
+       num_holes = sum(num1) - sum(num2)        ! number of "holes" we need to fill
+                                                ! to make the active dropelts in 
+                                                ! the "pdata" contiguous
+       i = nparcelsActive
+       j = 0
+       do np = 1, nparcelsActive
```
johnmauff: @sjsprecious It appears that you are looping over the entire collection of droplets looking for holes. Why are you not keeping track of the location of all the holes and then filling them in? Right now operation is O(nparcelsActive) versus what it could be which is O(num_holes)

src/comm.F
```diff
@@ -12947,8 +12951,8 @@ subroutine comm_droplet_value(holes_ind,num1,num2,pdata,pdata_neighbor)
       integer, intent(in) :: num2(num_nn)                                  ! Number of droplets that will enter the 
                                                                            ! current MPI region from each nearest 
                                                                            ! neighbor at different directions
-      real, intent(inout), dimension(nparcelsLocal,npvals) :: pdata         ! droplet information
-      integer, intent(in), dimension(nparcelsLocal) :: pdata_neighbor       ! array to store the new MPI region 
+      real, intent(inout), dimension(nparcelsLocal,npvals) :: pdata        ! droplet information
+      integer, intent(in), dimension(nparcelsActive) :: pdata_neighbor     ! array to store the new MPI region
```
johnmauff: Could you clarify if nparcelsActive is really the actual number of active particles or the maximum number of total particles?  If it is the former then why would you dimension arrays a size the should change during execution.

sjsprecious: `nparcelsActive` is the actual number of active droplets per MPI rank and it varies every time step. I think it is fine (and better) to specify the dimension explicitly as long as we update `nparcelsActive` correctly after the call to `droplet_driver`.

src/droplet.F
sjsprecious: Thanks for catching this inefficient implementation, @johnmauff ! I have revised the implementation based on your suggestions and the verification results are the same with the latest changes.

src/comm.F
sjsprecious: @johnmauff , I have updated the array dimension to a fixed length as we discussed.

src/droplet.F
```diff
@@ -420,101 +387,18 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
 
       ELSEIF (idropmethod.eq.1) THEN  !Implicit backward Euler for the droplet integration
```
johnmauff, line 389: @sjsprecious Right now it appears that if the user selects a different droplet time integration method the new MPI implementation will fail right?  I think we should be adding the neighbor array calculation to the rk2_integration time integration method.

src/droplet.F
```diff
@@ -704,18 +549,59 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
 
     call comm_droplet_number(holes_ind,num1,num2,pdata, &
                              pdata_locind,pdata_neighbor)
+    if ( sum(num1) .ne. 0 ) allocate(leave_ind(sum(num1)))
```
johnmauff: @sjsprecious 
The names of the num{1,2} arrays do not convey what they correspond to in the code.  You also use the calculation sum(num1) and sum(num2) quite frequently.  To make the code more readable I would suggest that you rename these arrays:  num1:= Depart, num2:= Arrive, and also define the new variables numDepart = sum(Depart), numArrive=sum(Arrive)

src/droplet.F
```diff
@@ -704,18 +549,59 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+    call comm_droplet_value(holes_ind,leave_ind,num1,num2,pdata,pdata_neighbor)
+
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+    ! Step 5: Compress the new "pdata" to make sure that all the active    ! 
+    !         droplets are contiguous (no "holes" between "pdata" slots)   !
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+    ! there will be remaining "holes" in the "pdata" only when more
+    ! droplets leave rather than enter the current MPI region
+    if ( sum(num2) .lt. sum(num1) ) then
```
johnmauff: @sjsprecious I am concerned about what you have done with respect to the hole-filling logic.  In particular, it has been placed in two locations.  Once here where the number of arriving droplets is less than the number of departing, and the other is in the comm.F file. Also, I don't think that the update_new_droplet() subroutine really does not belong in the comm.F file. I view this placement as just very confusing for long-term code maintainability.  We probably need a more involved discussion on this topic.

johnmauff (changes requested):

src/droplet.F
sjsprecious: @johnmauff I talked with David once about this issue and he commented that the RK2 solver should be removed sometime in the future. Thus we probably do not want to change it anyway.

src/droplet.F
sjsprecious: Thanks for the suggestion. Done.

johnmauff: Thanks this looks good and makes the code much easier to understand

src/comm.F
johnmauff: Thanks for the code modificaiton.

src/droplet.F
johnmauff: Great, thanks for reducing the order of the loop.

johnmauff (approved):
@sjsprecious Thanks for all the changes.  This PR now looks good.

src/droplet.F
johnmauff: Thanks for moving the hole-filling logic code to a single location in the code

PR #85, opened 2022-11-02
Runscripts and namelists for the PASC23 paper.

johnmauff:
This PR adds several namelist files that can be used to collect data for a PASC23 paper submission.  Note in addition to the namelists files several runscripts are also provided which can be used on Gust to submit jobs.


run/namelist.figureCD.cpu
```diff
@@ -0,0 +1,363 @@
...
+ irandp    =  1,
+ ibalance  =  0,
+ iorigin   =  1,
+ axisymm   =  0,
+ imove     =  1,
+ iptra     =  0,
+ npt       =  1,
+ pdtra     =  1,
+ iprcl     =  1,
+ nparcels  =  8000000,
```
sjsprecious, line 85: As we discussed, we may want to vary the number of droplets for a scaling test for Figure C & D. What `nparcels` shall I examine besides the value here?

run/namelist.figureCD.gpu
```diff
@@ -0,0 +1,363 @@
...
+ irandp    =  1,
+ ibalance  =  0,
+ iorigin   =  1,
+ axisymm   =  0,
+ imove     =  1,
+ iptra     =  0,
+ npt       =  1,
+ pdtra     =  1,
+ iprcl     =  1,
+ nparcels  =  8000000,
```
sjsprecious, line 85: The same question to the CPU namelist for Figure C & D.

run/gust.ngpu8.sh
```diff
@@ -0,0 +1,28 @@
...
+module purge
+module load nvhpc cuda cray-mpich
+
+export MPICH_GPU_SUPPORT_ENABLED=1
+#export CUDA_VISIBLE_DEVICES=$PMI_LOCAL_RANK
+export MPICH_GPU_MANAGED_MEMORY_SUPPORT_ENABLED=1
+#export NVCOMPILER_ACC_NOTIFY=16
+env 
+#mpiexec -n 8 -ppn 4 ./setdev.sh ./cm1.exe --namelist namelist.figureE.gpu >& gpu=8.figureE.512x512.log
+mpiexec -n 8 -ppn 4 ./setdev.sh ./cpu.exe --namelist namelist.figureD.gpu >& gpu=8.figureD.256.log
```
sjsprecious: Couple of comments:
- We may want to use the global script `get_local_rank` on Gust rather than a self-defined script here.
- Change `cpu.exe` to `gpu.exe`.
- Remove the `=` sign between `gpu` and `8`. Otherwise there is an annoying single quote in the output filename.

run/gust.node8.sh
```diff
@@ -0,0 +1,17 @@
...
+
+
+export PALS_NRANKS=1024
+export PALS_PPN=128
+export PALS_DEPTH=1
+export PALS_CPU_BIND=depth
+
+#mpiexec ./cm1.exe --namelist namelist.figureE.cpu >& node=8.figureE.512x512.log
+#mpiexec ./cpu.exe --namelist namelist.figureD.cpu >& node=8.figureD.256.log
+mpiexec ./cpu.exe --namelist namelist.figureA.cpu >& node=8.figureA.256.log
```
sjsprecious: Which compiler did you use for the CPU build/run? I tried the `nvhpc` compiler but the execution failed. Probably we need to load the corresponding modules here as well, similar to the GPU script.

sjsprecious (changes requested):
Thanks @johnmauff . This PR looks good to me in general but I do have a few comments. Based on my test, the GPU script works on Gust but the CPU script does not. May need additional clarification in your scripts.

run/namelist.figureCD.cpu
johnmauff: Your choice.  I would encourage you to collect data for different values and see if any interesting story emerges.  I selected 8M on 8 nodes, because the cost to run can be pretty large if it is increase any more.

run/namelist.figureCD.gpu
johnmauff: Again your choice.  The 40GB memory space of the A100 on Gust will support up to 4.5M per GPU, or 36M for 8 GPUs.  So that is the upper limit.

run/gust.ngpu8.sh
johnmauff: I am not aware of any get_local_rank script on Gust.  I am using the setdev.sh because that is what Brian V. and Rory recommend.

run/gust.node8.sh
johnmauff: I am using the standard environment and the nvhpc/22.7 compiler.  What kind of execution failure are you getting? Note that you will need to explicitly remove any cuda module from your environment otherwise it will give a runtime failure.

run/namelist.figureCD.cpu
sjsprecious: Gotcha!

run/gust.ngpu8.sh
sjsprecious: I was aware of this script on 10/13 so probably it was later than when you were testing CM1 on Gust. But since we are mapping one MPI rank per GPU, staying with your script may be ok.

run/gust.node8.sh
sjsprecious: Here is the error message from the output log file:
**file-transfer RPC timeout after 120s, received 0 of 1 responses**.

sjsprecious: I am using `nvhpc/22.7` as well and I do not load any CUDA module for the CPU run.

sjsprecious: Aha, I see your point now. I need to do "module unload cuda" in the run script even if it is for a CPU run. Now the CPU case is running on Gust.

sjsprecious: I would suggest to putting the following commands to your CPU run script:
```
module purge
module load nvhpc
module load cray-mpich
module load ncarcompilers
module unload cuda
```

sjsprecious (approved):

PR #88, opened 2022-12-13
Fix the race condition of tendency calculations on the GPU

sjsprecious:
This PR fixes the race condition of tendency calculations on the GPU by using the **!$acc atomic update** directive.

The verification result below shows that with this fix, the tendency calculation is done correctly on the GPU now.

- The verification result of metric variables using the existing `gpu-opt` branch (tendency is calculated on the CPU):
Identical variables: 44
variables with red_diff > 1.e-6: 29
variables with red_diff <= 1.e-6: 13
four variables with largest error: KSVMAX, KSHMAX, VOR2KM, VOR1KM

- The verification result of metric variables using the changes in this PR (tendency is calculated on the GPU):
Identical variables: 44
variables with red_diff > 1.e-6: 29
variables with red_diff <= 1.e-6: 13
four variables with largest error: KSVMAX, KSHMAX, VOR2KM, VOR1KM

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -96,12 +96,9 @@
       integer :: nrkp
       real :: dt2,uu1,vv1,ww1
       real :: sig3d,th_tmp,rand
-      real, dimension(nparcelsLocal) :: dvpdt1,dvpdt2,dvpdt3,drpdt,dtpdt,dmpdt, &
-                                        rp0,taup0,rhop0, &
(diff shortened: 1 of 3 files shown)


johnmauff (approved):
Thanks for fixing this race condition.  These fixes look good.

sherimickelson (approved):

PR #89, opened 2022-12-15
Fix a potential compiler bug for CUDA-aware MPI call

sjsprecious:
This PR fixes a potential compiler bug while invoking the `mpi_allreduce` for communication.

**Issue:**
- It turns out that the `!$acc update host` and `!$acc update device` directives around an `MPI_ALLREDUCE` call do not work properly and produce a wrong result in a GPU run unless the managed memory is used.

**Solution:**
- use the "!$acc host_data use_device" directive
- specify the array bounds explicitly in the "!$acc update host" and "!$acc update device" directives

Both solutions work so far and we decide to go with the first option since it enables the CUDA-aware MPI communication. No managed memory is needed for the new GPU run.

**Verification result:**

GPU run with managed memory (from https://github.com/george-bryan/CM1/pull/88):
- Identical variables: 44
- variables with red_diff > 1.e-6: 29
- variables with red_diff <= 1.e-6: 13
- four variables with largest error: KSVMAX, KSHMAX, VOR2KM, VOR1KM

GPU run without managed memory:
- Identical variables: 44
- variables with red_diff > 1.e-6: 29
- variables with red_diff <= 1.e-6: 13
- four variables with largest error: KSVMAX, KSHMAX, VOR2KM, VOR1KM

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -45,7 +45,6 @@ DM            =
 OMP           =
 DP            =
 OPTS          =
-OPTS_M        =
 CPP           =
 OUTPUTINC     =
 OUTPUTLIB     =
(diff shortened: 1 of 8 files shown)


src/Makefile
```diff
@@ -116,7 +116,7 @@ ifeq (${FC},nvfortran)
                 OPTS_M += -acc -gpu=cc80,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
             else
                 OPTS += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
-                OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform,managed -Mpcast -Minfo=accel
+                OPTS_M += -acc -gpu=cc70,lineinfo,nofma,math_uniform -Mpcast -Minfo=accel
```
johnmauff: Do we want to maintain an OPTS_M compiler flag in the Makefile now that we do not need the use of Managed memory?

johnmauff (commented):
Thanks for submitting this PR.  It looks good with the potential exception of the issue in the Makefile.

johnmauff:
@sjsprecious, I noticed that there are other '!$acc update host()' that are associated with MPI reduction calls.  For example in solve1.F and solve2.F. Is there a particular reason why these were not modified?

src/Makefile
sjsprecious: Thanks @johnmauff. This is a good point. I keep it here in case we want to apply separate flags to different files in the future. But it does seem unnecessary by now since we no longer need the managed memory option. Shall we just remove it?

sjsprecious:
> @sjsprecious, I noticed that there are other '!$acc update host()' that are associated with MPI reduction calls. For example in solve1.F and solve2.F. Is there a particular reason why these were not modified?

Thanks for catching them. I do not modify them because it is a `mpi_iallreduce` call and thus skips from my `grep` search. Let me go through the code again and eliminate the potentially buggy `!$acc update host` directive.

src/Makefile
sjsprecious: The Makefile has been updated to remove the `OPTS_M` option.

sjsprecious:
@johnmauff I have removed additional `!$acc update host()` directives in other source code. The verification results remain unchanged. Let me know if you have further questions.

johnmauff:
@sjsprecious Thanks for the updates.  It now looks good.

sherimickelson (approved):
Thanks @sjsprecious for adding in these changes

PR #90, opened 2022-12-22
Add gpudirect

johnmauff:
This commit eliminates a number of _GPUDIRECT if-def blocks.  The elimination indicates that GPU-direct is now by default in a number of subroutines.   Note that four subroutines in comm.F still do not have GPU-direct enabled by default as they still have bugs that have not yet been resolved.  The correctness statistics are unchanged 

44 stat variables are identical.
29 stat variables show a mean relative difference > 1e-06
13 stat variables show a mean relative difference <= 1e-06

### DIFF
--- a/src/comm.F
+++ b/src/comm.F
@@ -8376,9 +8376,6 @@
       !$acc declare present(w) &
       !$acc present(ww1,ww2,we1,we2,ws1,ws2,wn1,wn2)
 
-#ifndef _GPUDIRECT17
-      if(Debug) print *,'comm_1w_start_GPU:'
-#endif
       nr = 0
 !-----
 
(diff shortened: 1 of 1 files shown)


sherimickelson (approved):
Thanks, @johnmauff !  Looks good to me.

sjsprecious (approved):
Thanks @johnmauff for submitting this PR. It looks good to me and I could confirm that the verification results do not change against the **ifort** baseline.

One clarification question: do we still need to keep the CPU version of some subroutines like **getcorner3**, **comm_3s_end**, etc.

johnmauff:
@sjsprecious thanks for the review.  Currently, to the best of my knowledge, we still need to keep the duplicate subroutines in the comm.F file.  I did learn something last week that gives me an idea about how we can eliminate the CPU versions of the files in comm.F.   I have not yet tested the idea yet though.

Issue #94, opened 2023-01-30
Restart fails on GPU without droplets


johnmauff:

johnmauff:
PR #95 addresses this issue

PR #95, opened 2023-01-30
Added if_present clause in restart.F

johnmauff:
A number of variables present on the restart file have not been currently moved to the GPU.  This is a result of not GPU-enabling all of CM1 but rather just the sections of the model that are necessary for the ASD simulation.  The addition of the if_present avoids updating the version of variables on the host unless a version exists on the device.  The correctness statistics are unchanged.  This PR addresses issue #94

### DIFF
--- a/src/restart.F
+++ b/src/restart.F
@@ -4591,7 +4591,6 @@
 #ifdef MPI
     rf2:  IF( restart_filetype.eq.3 )THEN
 
-      print *,'writer: before writer2'
       call    writer2(numi,numj,numk1,numk2,nxr,nyr,var,aname,           &
                       ni,nj,ngxy,myid,numprocs,nodex,nodey,orec,nfile,   &
                       ncid,time_index,restart_format,restart_filetype,myi1p,myi2p,myj1p,myj2p,   &
--- a/src/restart.F
+++ b/src/restart.F
@@ -573,91 +573,91 @@
 ! standard 2D:
 
       n = 1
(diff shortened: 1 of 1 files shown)


src/restart.F
```diff
@@ -4591,6 +4591,7 @@ subroutine writer(numi,numj,numk1,numk2,nxr,nyr,var,aname,           &
 #ifdef MPI
     rf2:  IF( restart_filetype.eq.3 )THEN
 
+      print *,'writer: before writer2'
```
sherimickelson: @johnmauff Is this line still needed?

johnmauff: Thanks for that catch.  I just deleted it.

sjsprecious:
Thanks @johnmauff for addressing this issue. Do I understand it correctly that with these changes, the restart functionality works without droplet microphysics enabled? Are the results BFB?

johnmauff:
@sjsprecious I have actually not checked B4B yet.  I will be doing that next.  These modifications are necessary for the GPU version of the code to actually run to completion.  Without them, it fails with a variable, not on device error.  I do anticipate the need for addition code modifications.

sjsprecious:
Thanks for the clarification. That is clear to me now.

sjsprecious (approved):

src/restart.F
sherimickelson: Thanks @johnmauff

sherimickelson (approved):

PR #96, opened 2023-02-03
David physics 2

drichte2:
Adds the ability of droplets to independently fall out and be injected into the domain. "nparcelsActiveMax" is the maximum value that "nparcelsActive" can become, which is always less than "nparcelsLocal" by the buffer size. Also fixed a major performance issue due to single precision issues in the nonlinear, iterative droplet solver.


src/cm1.F
```diff
@@ -2259,11 +2259,12 @@ program cm1
       allocate( reqt(d3t) )
       reqt = 0
 
+      nparcelsActiveMax = 1e6   !The total allowable number of particles on each processor
 #ifdef MPI
-      padding = max(pdata_buffer*nparcels,100.0)
-      nparcelsLocal = ceiling((nparcels*1.0/numprocs)+padding)
+      padding = max(pdata_buffer*nparcelsActiveMax,100.0)
```
sjsprecious: By making this change, the value of `padding` is now independent of the `nparcels` value. I think a sanity check is needed to make sure that `nparcels/numprocs` is always smaller than `nparcelsActiveMax`.

src/Makefile
```diff
@@ -156,7 +156,7 @@ endif
 ifeq (${FC},ifort)
     # replace -xHost with -march=core-avx2 to make sure the same code builds on
     # both Intel and AMD CPUs 
-    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
```
sjsprecious: I think this flag won't work on Gust's CPU. Better to revert this change.

src/droplet.F
```diff
@@ -126,9 +133,20 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
       integer :: tmp_flag
 #endif
 
+      double precision :: t_s,t_f,t_findloc,t_interp,t_BEint,t_feed,t_s1,t_f1,t_BEprep,t_radsolver,t_gauss,t_LV,t_neighbor
```
sjsprecious: Do we need to keep these timers during our ASD runs or are they just for your own diagnostic?

src/droplet.F
```diff
@@ -126,9 +133,20 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+
+
+      t_s = mpi_wtime()
+
+      !!! DHR: From what I can tell this is useless -- it only resets seed for the intrinsic random_number routine
```
sjsprecious: I agree and I thought you would remove these calls and take the `randomNumbers` from the `init3d` subroutine to pull new random numbers.

src/droplet.F
```diff
@@ -302,9 +319,27 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+    t_interp = 0.0
+    t_BEint = 0.0
+    t_feed = 0.0
+
+    t_BEprep = 0.0
+    t_radsolver = 0.0
+    t_gauss = 0.0
+    t_LV = 0.0
+    t_neighbor = 0.0
+
```
sjsprecious: These changes will break the GPU code. Need to be moved up to line 311.

src/droplet.F
```diff
@@ -445,8 +468,73 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+         pdata_locind(holes_fallout_ind(np),:) = pdata_locind(nparcelsActive,:)
+         pdata_neighbor(holes_fallout_ind(np)) = pdata_neighbor(nparcelsActive)
+         nparcelsActive = nparcelsActive-1
+      end if
+
+   end do
+
+   deallocate(holes_fallout_ind)
+
+   call mpi_allreduce(nparcelsActive,np_tmp,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
```
sjsprecious: why do we need an `mpi_allreduce` call here?

src/droplet.F
```diff
@@ -527,23 +617,49 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+! Final thing: Inject droplets which will be active on the next call of droplet_driver
+! Requires inputting the current 10-meter wind speed
+     
+     s10_xym = 0.0
+     do j=1,nj
+     do i=1,ni
+        s10_xym = s10_xym + s10(i,j)
+     end do
+     end do
+     call mpi_allreduce(mpi_in_place,s10_xym,1,mpi_real,mpi_sum,mpi_comm_world,ierr)
```
sjsprecious: Again, why do we need an `mpi_allreduce` call here?

src/droplet.F
```diff
@@ -645,34 +764,48 @@ subroutine BE_integration(dt,np,pdata,pdata_locind,xf,yf,x3d,y3d, &
...
-      if ( (rt_zeros(1) .ne. rt_zeros(1)) .OR. &
-           (rt_zeros(2) .ne. rt_zeros(2)) .OR. &
-           (rt_zeros(2) .lt. 0) ) then
-         !If this happens keep radius unchanged and temp equal to surrounding
-         rt_zeros(1) = 1.0
-         rt_zeros(2) = tval/pdata(np,prtp)
+      if (flag==1)  then
+
+       num1000 = num1000+1 
+
```
sjsprecious, line 747: The `if` checks for `rt_zeros` are removed, is it intended or not?

src/droplet.F
```diff
@@ -689,11 +822,12 @@ subroutine BE_integration(dt,np,pdata,pdata_locind,xf,yf,x3d,y3d, &
...
-         !z3d = 1.0e-6  !Just temporary 
-         !pdata(np,pract) = -1.0  !Signal this as no longer alive
-         z3d = bot + (bot-z3d)
-         pdata(np,prvpz) = -pdata(np,prvpz)
+         z3d = 1.0e-6  !Just temporary 
+         pdata(np,pract) = -1.0  !Signal this as no longer alive
+         !z3d = bot + (bot-z3d)
+         !pdata(np,prvpz) = -pdata(np,prvpz)
          pdata_locind(np,3) = undefined_index
+         num_fallout = num_fallout + 1
```
sjsprecious: For my own reference: `num_fallout` needs to be put on the `reduction` clause for GPU code.

src/droplet.F
```diff
@@ -2270,6 +2419,196 @@ subroutine ie_vrt_nd(diffnorm,tempr,tempt,rt_output,dt_nondim,taup_scale,prsval,
...
+      my_reintro = 10
+
+   if (myid==0) write(*,'(a15,i,6e15.6)')  'DROP INJECTION:',my_reintro,drop_mult,u10,totdrops,xrange,yrange,dt
+
+      if (nparcelsActive+my_reintro .lt. nparcelsActiveMax) then   !Only add particles to pdata if we're still under nparcelsActiveMax
+
+      do np=nparcelsActive+1,nparcelsActive+my_reintro
+
+         rand = getRandomReal(randomNumbers)
+         a = totdrops*getRandomReal(randomNumbers)
```
sjsprecious: should it be `a = totdrops * rand`?

src/droplet.F
```diff
@@ -2270,6 +2419,196 @@ subroutine ie_vrt_nd(diffnorm,tempr,tempt,rt_output,dt_nondim,taup_scale,prsval,
...
+
+         do i = 1,100
+            if ((a .gt. dFssum(i)) .and. (a .le. dFssum(i+1))) then
+               rand = getRandomReal(randomNumbers)
+               rad_init = binsdata(i) + rand*(binsdata(i+1)-binsdata(i))
+            endif
+         enddo
```
sjsprecious, line 2607: Can you add some comments about what this loop is doing here?

sjsprecious (changes requested):
Thanks @drichte2 for submitting this PR. I have a few questions and hope you can clarify them before I can work on the fix for the GPU code.

sjsprecious:
@drichte2 : also how did you verify the correctness of your new code?

src/droplet.F
```diff
@@ -445,8 +468,73 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+    if (myid==0) write(*,*) 'TIME BEprep:',t_BEprep
+    if (myid==0) write(*,*) 'TIME radsolver:',t_radsolver
+    if (myid==0) write(*,*) 'TIME gauss:',t_gauss
+    if (myid==0) write(*,*) 'TIME LV:',t_LV
+    if (myid==0) write(*,*) 'TIME neighbor:',t_neighbor
+
+    if (myid==0) write(*,*) 'TIME feed:',t_feed
+
+    call mpi_allreduce(mpi_in_place,num100,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
+    call mpi_allreduce(mpi_in_place,num1000,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
```
johnmauff: Why are we performing two calls to MPI_Allreduce? A better solution would be to perform a single MPI_Allreduce call on a vector of length two.

src/init3d.F
```diff
@@ -423,7 +423,8 @@ subroutine init3d(xh,rxh,uh,ruh,xf,rxf,uf,ruf,yh,vh,rvh,yf,vf,rvf,  &
           rand = getRandomReal(randomNumbers)
           tmpy = rand*maxy
           rand = getRandomReal(randomNumbers)
-          tmpz = rand*maxz
+          !tmpz = rand*maxz
+          tmpz = rand*8.0
```
johnmauff: Could you explain why the variable maxz was replaced with a magic number 8?

src/misclibs.F
```diff
@@ -2986,7 +2986,7 @@ subroutine calccfl(nstat,rstat,dt,acfl,uh,vh,mh,ua,va,wa,writeit)
 
 !!!      cflmax = fmax
 
-      print *,'calccfl: fmax is: ',fmax
+!      print *,'calccfl: fmax is: ',fmax
```
johnmauff: Lets just delete this print statement and the next one on line 3170.t

src/cm1.F
drichte2: Good point. The padding should actually be a percentage of `nparcelsActiveMax`, since otherwise the padding would need to be adjusted throughout the simulation (which doesn't seem ideal/possible). So maybe I can change to `nparcelsActiveMax/numprocs` instead?

src/Makefile
drichte2: Yes, this was a mistake since it doesn't work on my local machine. Will change for the pull request

src/cm1.F
johnmauff: I am also not comfortable with this change both for the reason that Jian mentioned and also for the fact that there are now two places where the number of droplets is set.  Once in the namelist and then once in this file.  That is a recipe for confusion.

src/droplet.F
drichte2: No, I meant to take these out. I have removed them from the pull request

src/droplet.F
drichte2: Okay I have removed this

src/droplet.F
drichte2: These were all for my timing tests, I have removed them

src/droplet.F
drichte2: I have moved this down below. At the moment it's just for a writeout statement that lets me know how many particles are currently in the system (variable `np_tmp`). This can be removed later

src/droplet.F
drichte2: All of the processors need to know the value of `s10_xym`. This is the 10-meter wind speed that goes into the droplet injection routine and determines how many droplets need to be produced.

src/droplet.F
drichte2: Yes, this was intended. The checks on `rt_zeros` are now above, and they set `flag = 1` which triggers this code.

src/droplet.F
drichte2: Yes, my mistake

src/droplet.F
drichte2: I have added some comments. Let me know if they don't explain well

drichte2:
Correctness was done by ensuring the droplet solver for a fixed background value-- this was how I had done this the first time.

src/cm1.F
sjsprecious: Thanks @drichte2 . It is up to you to decide the calculation of `padding`. My point is that we need to check the value of `nparcels` here. For example, if I set `nparcels = 1 billion`, each MPI rank will have more than `nparcelsActiveMax` droplets and ideally the program should stop with an error message. This is missing from the current implementation.

src/droplet.F
sjsprecious: But this means all the processors will know the maximum 10-m wind speed over the whole domain. Should it be enough for each processor to know the 10-m wind speed of its own sub-domain?

sjsprecious:
> Correctness was done by ensuring the droplet solver for a fixed background value-- this was how I had done this the first time.

Thanks @drichte2 . It is still unclear to me how to verify the correctness quantitatively, like which statistic output you are looking at. We should have an offline discussion about this in detail.

src/droplet.F
drichte2: Strictly speaking, it should really be the domain-wide average. In practice, it probably wouldn't make a big difference if we did it locally, but this would introduce a dependence on `numprocs` since different processor numbers would lead to different areas that this averaging is being done over.

sjsprecious: thanks for the clarification.

johnmauff (changes requested):

src/cm1.F
drichte2: I just added a quick check to see if `nparcels` exceeds `nparcelsActiveMax`. @johnmauff , it's not setting the same thing twice. `nparcels` is the number of droplets to start the simulation with. `nparcelsActiveMax` is a maximum upper limit on the number of parcels that can ever be active at the same time. This would be something that should only be changed with architecture/memory limitations

src/init3d.F
drichte2: This is the maximum height that the droplets will be "injected". Problem-specific, and has to do with things like the (unresolved) wave height, etc. Using `maxz` here is like saying that a new droplet could magically appear anywhere in the domain, when in reality it should start near the bottom.

src/misclibs.F
drichte2: Done

src/droplet.F
drichte2: Changed

src/cm1.F
johnmauff: @drichte2  Historically nparcels was the total number of particles.  I think the current use is a bit confusing and error prone.   Maybe we need to have two namelist variables:  
 
nparcelsInit:  Initial number of particles
nparcelsMax:  Maximum number of particles.  

Might want to get @george-bryan input as we are proposing a namelist change.

src/init3d.F
johnmauff: Ok that makes sense.  We probably need to add this as a namelist parameter.  The current use of a number that is not documented makes the code more fragile than necessary.

src/cm1.F
drichte2: I agree with the last part -- I was trying to do this in a way that didn't mess with the namelist file. I'd be fine adding it at some point, but didn't want to mess up everyone's namelists

src/init3d.F
drichte2: Agreed -- see my other comment about adding things to the namelist. That would make sense, but didn't want to mess with that part

src/param.F
```diff
@@ -297,7 +297,7 @@ subroutine param(namelist_input,                                  &
       call bcast_int(npt)
       call bcast_int(pdtra)
       call bcast_int(iprcl)
-      call bcast_int(nparcels) ! DEVICE
+      call bcast_int(nparcelsInit) ! DEVICE
```
johnmauff, line 300: @sjsprecious Where is the corresponding bcast_int() call for nparcelsMax?

src/cm1.F
```diff
@@ -2259,15 +2259,18 @@ program cm1
       allocate( reqt(d3t) )
       reqt = 0
 
+      if ( nparcelsInit .gt. nparcelsMax ) then
+          if ( myid .eq. 0 ) print *,' WARNING: nparcelsInit exceeds allowed max. Check nparcelsMax. Exiting...'
```
johnmauff, line 2263: @sjsprecious  I am concerned that you have removed the necessary padding for setting the value for nparcelsLocal.  How do you prevent code failure if the number of incoming particles exceeds nparcelsLocal.

src/param.F
sjsprecious: @johnmauff thanks for catching this! Just added it.

src/cm1.F
sjsprecious: @johnmauff with the new namelist variable `nparcelsMax`, `nparcelsLocal` is now calculated as `nparcelsMax/numproc` and I think we do not need the `padding` variable now. In the `init3d.F` subroutine, there is a check if the number of incoming particles exceeds `nparcelsLocal` and an error will issue accordingly.

johnmauff: @sjsprecious I think the padding still needs to be present.  Right now the code could fail unpredictably later in the simulation if the number of incoming droplets is sufficiently large.  Another option to test this functionality would be to set nparcelsMax==nparcelsInit.  The code should then fail for the first MPI rank that receives more incoming droplets than depart.

sjsprecious: @johnmauff thanks for the detailed explanation. That makes sense. However, if we use the variable `padding` to increase the size of `nparcelsLocal`, how could we ensure that the total droplets after injection will not exceed `nparcelsMax` at each time step?

johnmauff:
@sjsprecious, @drichte2  @sherimickelson I just took a close look at these changes and I have an increasingly large number of concerns.  For example, the droplet code is mostly missing any comments, there has been code added that will prevent the code from compiling without defining MPI and there is still no clarity regarding how to set various maximum # of particles.  There is also missing OpenACC support for several newly added loops.  I am concerned that allowing this PR will immediately break a number of execution modes for CM1 that currently work.  Ideally, I would like to run some performance measurements on David's new additions, but I don't understand how I can configure the code so that it runs on a constant particle # to allow comparison with the ISC paper results.  I wonder if we need to discuss these code changes synchronously instead of through GitHub.

sjsprecious:
Thanks @johnmauff . I agree that we may need another meeting to address these issues. I also find that with the changes here, the restart run does not agree with the initial run anymore (though the difference is not that large, e.g., total droplet number).

sjsprecious:
@johnmauff In terms of conserving the total droplet number, I just pushed a new change and now by setting `use_wall_bc = .true.`, we will have a constant droplet number through the simulation.

johnmauff (approved):

src/droplet.F
```diff
@@ -400,53 +416,72 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+
+    if ( myid == 0 ) then
+       call mpi_reduce(mpi_in_place,numvec,2,mpi_int,mpi_sum,0,mpi_comm_world,ierr)
+    else
+       call mpi_reduce(numvec,numvec,2,mpi_int,mpi_sum,0,mpi_comm_world,ierr)
+    end if
+
+#endif
+
+    if (myid==0) write(*,*) 'num100,num1000 = ',numvec(1),numvec(2)
```
sjsprecious, line 449: Are these for debug purpose only or do we need to include them in a production run?

src/droplet.F
```diff
@@ -515,35 +550,61 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+#endif
+      s10_xym = s10_xym/real(nx*ny)
+      if (myid==0) write(*,*) 'DHR U10',s10_xym
+      call droplet_injection(pdata,xf,yf,dt,s10_xym,randomNumbers)
+#ifdef MPI
+      call mpi_allreduce(nparcelsLocalActive,np_tmp,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
+#endif
+      if (myid==0) then 
+         write(*,'(a8,3i)')  'DHR1:',myid,np_tmp,nparcelsLocalActive
+      end if
```
sjsprecious: Do we need to keep this print statement?

src/droplet.F
```diff
@@ -515,35 +550,61 @@ subroutine droplet_driver(dt,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,    &
...
+      do j=1,nj
+         do i=1,ni
+            s10_xym = s10_xym + s10(i,j)
+         end do
+      end do
+#ifdef MPI
+      call mpi_allreduce(mpi_in_place,s10_xym,1,mpi_real,mpi_sum,mpi_comm_world,ierr)
+#endif
+      s10_xym = s10_xym/real(nx*ny)
+      if (myid==0) write(*,*) 'DHR U10',s10_xym
```
sjsprecious: Do we need to keep this print statement?

src/droplet.F
drichte2: I'll remove the print statement in the next PR, but the calculations need to happen. I need to send them to the droplet_diag

src/droplet.F
drichte2: No, not in the long run. I'll remove it for the next PR

src/droplet.F
drichte2: Same -- not in the long run, but it's helpful now as I set up the ASD case. I'll remove for the next PR

sjsprecious (approved):

sjsprecious:
Somehow I could not revert some commits in this PR so I end up with opening a separate PR for all the approved changes (available at https://github.com/george-bryan/CM1/pull/100) and closing this PR without merging.

PR #97, opened 2023-02-08
Fix GPU restart

johnmauff:
Restart runs on the GPU are now B4B when comparing the Metric terms when the droplet capability is deactivated.


sherimickelson (approved):
Thanks @johnmauff

run/namelist.restB
```diff
@@ -0,0 +1,363 @@
...
+
+ &param1
+ dx     =    40.0,
+ dy     =    20.0,
+ dz     =    10.0,
+ dtl    =   0.125,
+ timax  = 120.0,
+ run_time =  -999.9,
+ tapfrq =   900.0,
+ rstfrq = 10800.0,
```
sjsprecious, line 21: should this `rstfrq` be set to `60` here?

sjsprecious:
@johnmauff : thanks for submitting this PR. Could you please provide more details about how to use these namelists?

run/namelist.restB
johnmauff: You could set this to 60 if you want but it is really not crtical because the restart file that is written by the namelist.restB run is never used.

johnmauff:
I have added three namelist files to this PR.  They are:

namelist.restA:  initial run for 60 seconds and write a restart
namelist.restB:  restart run from 60 seconds and run to 120 seconds
namelist.restC: Initial run for 120 seconds

I use them in the following way 

mpirun cm1.exe --namelist namelist.restA
mpirun cm1.exe --namelist namelist.restB
mpirun cm1.exe --namelist namelist.restC

src/restart.F
```diff
@@ -3646,8 +3653,7 @@ subroutine read_restart(nstep,srec,sirec,urec,vrec,wrec,nrec,mrec,prec,      &
         print *,'  nvar_parcels = ',nvar
       endif
 
-#ifdef MPI
-      call MPI_BCAST(nvar,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
+      call bcast_int(nvar)
 #endif
```
sjsprecious: This `#endif` should be removed.

sjsprecious (changes requested):

src/restart.F
johnmauff: Just fixed it.

sjsprecious:
@johnmauff : I have tested the `fix_restart` branch with both CPU and GPU runs. The CPU initial run and restart run seem to be BFB. However, the CPU and GPU restart runs (using the same restart files from the previous CPU run) do not agree with each other. Could you explain more about the statement of **Restart runs on the GPU are now B4B** in your PR's description?

johnmauff:
@sjsprecious  You can not mix CPU and GPU runs and still get B4B results.  Instead, compare the output of a restarted GPU run versus one that was not.  So look at metric values from GPU runs using namelists rest{A,B} and restC.  They should match B4B.

sjsprecious:
> @sjsprecious You can not mix CPU and GPU runs and still get B4B results. Instead, compare the output of a restarted GPU run versus one that was not. So look at metric values from GPU runs using namelists rest{A,B} and restC. They should match B4B.

Thanks @johnmauff for your clarification. I could confirm that the rest and init runs are BFB on CPU and GPU, respectively.

sjsprecious (approved):

PR #98, opened 2023-02-12
Fix the restart issue with droplet microphysics enabled

sjsprecious:
This PR introduces an I/O interface for the droplets based on the new MPI implementation and fixes the issues during a restart run with droplet microphysics enabled.

To be specific, the current `restart.F` code only outputs the x/y/z location of droplets to the restart file (i.e., the `ploc` variable), which are not enough for a restart run. Therefore, we output the whole `npvars` variables in the `pdata` to the restart file instead. Note that this will increase the memory usage when writing/reading a restart file.

The initial and restart runs indicate that the results are BFB either on CPU or GPU (need to remove the initial `stat::` and droplet diagnostic output first before using the verification script).


sherimickelson (approved):

johnmauff (approved):
@sjsprecious Thanks for fixing the droplet-based restart capability.

PR #99, opened 2023-02-20
Syntax errors for directives identify by CCE

johnmauff:
This PR resolves a number of issues with the OMP and ACC directives that were identified by the CCE compiler.  It also reduces the default level of compiler optimization level for the CCE compiler from O3 to O2 because the O3 level takes a very long time to compile.

### DIFF
--- a/src/adv_routines.F
+++ b/src/adv_routines.F
@@ -3384,8 +3384,7 @@
 
   IF( stag.eq.1 )THEN
 
-!$omp parallel do default(shared)   &
-!$omp private(i,j,k),
+    !$omp parallel do default(shared) private(i,j,k)
     !$acc parallel loop gang vector collapse(3) default(present)
     DO j=1,nj
       do k=3,nk-1
--- a/src/bc.F
+++ b/src/bc.F
@@ -1585,7 +1585,6 @@
             radbcs(i,k)=avgs
           enddo
         enddo
-        !$acc end parallel
       endif
 
       if(ibn.eq.1.and.nbc.eq.2)then
@@ -1605,7 +1604,6 @@
             radbcn(i,k)=avgn
           enddo
         enddo
-        !$acc end parallel
       endif
 
       if(timestats.ge.1) time_bc=time_bc+mytime()
@@ -1646,7 +1644,6 @@
         temout(k) = 0.0d0
         temin(k)  = 0.0d0
       enddo
-      !$acc end parallel
 
       if(wbc.eq.2.and.ibw.eq.1)then
         i=1
@@ -1694,7 +1691,6 @@
         fluxout = fluxout + temout(k)
         fluxin  = fluxin  + temin(k)
       enddo
-      !$acc end parallel
 
 #ifdef MPI
       tem=0.0d0
@@ -1721,7 +1717,6 @@
           endif
         enddo
         enddo
-        !$acc end parallel
       endif
 
       if(ebc.eq.2.and.ibe.eq.1)then
@@ -1736,7 +1731,6 @@
           endif
         enddo
         enddo
-        !$acc end parallel
       endif
 
       !$acc end data
@@ -1815,7 +1809,6 @@
         fluxout = fluxout + temout(k)
         fluxin  = fluxin  + temin(k)
       enddo
-      !$acc end parallel
 
 #ifdef MPI
       tem=0.0d0
--- a/src/sfcphys.F
+++ b/src/sfcphys.F
@@ -548,7 +548,7 @@
 
 
       !$acc declare &
-      !$acc present(psfc,tsk,th0,tha,qa,tsk,znt,za) &
+      !$acc present(psfc,tsk,th0,tha,qa,znt,za) &
       !$acc present(dsxy,s1,thflux,qvflux,rf,cd,hpbl) &
       !$acc present(qsfc,gz1oz0,wspd,br,hfx,qfx) &
       !$acc present(cda,psim,psih,zol,mol,fm,fh,th2,q2,t2)
--- a/src/turbtend.F
+++ b/src/turbtend.F
@@ -2938,10 +2938,10 @@
         enddo
         enddo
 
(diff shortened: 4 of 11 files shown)


src/adv_routines.F
```diff
@@ -3384,8 +3384,7 @@ subroutine vadv_flx3(stag,ix,jy,kz,c1,c2,rrw,dumz,a)
 
   IF( stag.eq.1 )THEN
 
-!$omp parallel do default(shared)   &
-!$omp private(i,j,k),
+    !$omp parallel do default(shared) private(i,j,k)
```
sjsprecious, line 3387: why is this a syntax error? It seems to me a stylish choice.

src/bc.F
```diff
@@ -1585,7 +1585,6 @@ subroutine radbcns4(rvf,radbcs,radbcn,v1,v2,dt)
             radbcs(i,k)=avgs
           enddo
         enddo
```
sjsprecious, line 1587: does't make sense to me. Why do we have to remove `!$acc end parallel`?

src/sfcphys.F
```diff
@@ -548,7 +548,7 @@ subroutine sfcdiags(tsk,thflux,qvflux,cd,ch,cq,u1,v1,s1,wspd,        &
 
 
       !$acc declare &
-      !$acc present(psfc,tsk,th0,tha,qa,tsk,znt,za) &
+      !$acc present(psfc,tsk,th0,tha,qa,znt,za) &
```
sjsprecious, line 551: interesting. Does `tsk` ever exist on the device? Why nvhpc compiler dose not catch this error?

src/turbtend.F
```diff
@@ -2938,10 +2938,10 @@ subroutine turbw(dt,xh,rxh,arh1,arh2,uh,xf,vh,mh,mf,rho,rf,gz,rgzu,rgzv,rds,sigm
         enddo
         enddo
 
-        !$acc parallel loop gang vector default(present) private(i,j)
+        !$acc parallel default(present)
```
sjsprecious, line 2941: There is a dependency of `k` for `dum3` calculation, so the `k` loop must be computed sequentially. I think we need to move the `!$acc parallel default(present)` directive down to the `i,j` loop. Or we could move the `k` loop inside the `i,j` loop and run it sequentially. What do you think? @johnmauff

sjsprecious (changes requested):

sjsprecious (changes requested):
Thanks @johnmauff for fixing these issues. Many changes related to `!$acc end parallel` directive actually do not make sense to me. Which OpenACC standard does the Cray compiler support? I wonder if we are fixing issues for an old OpenACC standard.

src/adv_routines.F
johnmauff: @sjsprecious, the presence of the comma after the priviate(i,j,k) clause is a syntax error

src/bc.F
johnmauff: @sjsprecious, there is no corresponding '$acc parallel' here, but rather a '$acc parallel loop'.  Apparently, nvhpc is not bothered by this but CCE is.

src/sfcphys.F
johnmauff: @sjsprecious the syntax error came from the fact that the 'tsk' variable is included twice in the declare present clause.

src/adv_routines.F
sjsprecious: Oh, thanks for pointing that out. Sorry that I did not catch it!

src/turbtend.F
johnmauff: @sjsprecious, I think the current setting for this loop will run the k loop in serial, however, I am not 100% certain of this.  Note that the previous version contains two 'parallel loop' clauses which were not supported by CCE and are not correct due to the k dependency.  Maybe this needs further discussion.

src/bc.F
sjsprecious: It should be ok to match `!$acc parallel loop` with `!$acc end parallel` since there is no `!$acc end loop`. I guess this issue is caused by the fact that CCE is supporting an old OpenACC standard?

src/sfcphys.F
sjsprecious: Oh, I see. Thanks for pointing it out.

src/turbtend.F
sjsprecious: My guess is that the `k` loop will be run redundantly in the current setting. Hopefully it is just a matter of efficiency not correctness. But thanks for finding this issue out.

src/bc.F
johnmauff: I don't know if it is standard issues or the nvhpc compiler that is more forgiving.  Regardless, I am hoping to eventually get the OpenACC version of CM1 to compile using both NVHPC and CCE compilers.

src/turbtend.F
johnmauff: @sjsprecious So this is not the only place where this OpenACC loop construct exists.  Why would it be less efficient?  I want all threads to execute the k loop followed by an assignment of each thread to its own value of i and j.

src/bc.F
sjsprecious: Ok, that sounds fine to me.

src/turbtend.F
sjsprecious: Hmm, after a second thought, I think you are correct. So my only concern here is whether the `k` loop is executed sequentially or in a random order. Maybe pushing `k` loop inside the `i,j` loops and adding `!$acc loop seq` to it is a safer way?

johnmauff: @sjsprecious Inverting the loop would potentially have an impact on CPU performance.  Also, a significant change like that might also impact correctness.  The goal of this PR was just to fix the syntax errors that were identified.

sjsprecious: I see and it makes sense.

sjsprecious (approved):

sherimickelson (approved):

PR #101, opened 2023-03-31
Fix the broken GPU code with droplet injection function

sjsprecious:
This PR fixes the broken GPU code due to the changes from this PR #100.

**CPU (intel, 2 MPI ranks) vs. GPU (nvhpc, 2 MPI ranks + 2 V100 GPUs)**

_Stat output comparision_
- Identical variables: 59
- variables with rel_diff > 1.e-6: 24
- variables with rel_diff <= 1.e-6: 3
- four variables with largest error: PPMIN, PPMAX, DIVMIN, DIVMAX

_Droplet 0th-order diagnostic comparision_
- Identical variables: 12
- variable with rel_diff > 1.e-6: 0
- variable with rel_diff <= 1.e-6: 0

_Droplet 1st-order diagnostic comparision_
- Identical variables: 512
- variable with rel_diff > 1.e-6: 0
- variable with rel_diff <= 1.e-6: 0

--------------------------------------------------------------------------------

**Note** that if we use the droplet injection function, the total number of droplets at the end of simulation varies when we use different number of MPI ranks. Thus in the verification test, we need to use the same number of MPI ranks for both CPU and GPU runs.

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -563,6 +563,7 @@
 ! Requires inputting the current 10-meter wind speed
    
    if ( .not. use_wall_bc ) then  
+      s10_xym = 0.0
       !$acc parallel loop gang vector reduction(+:s10_xym)
       do j=1,nj
          do i=1,ni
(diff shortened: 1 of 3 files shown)


johnmauff (approved):
Thanks for the PR.  This looks great.

PR #103, opened 2023-04-06
Sub-PR 1

drichte2:
Trying to break up the PR with various milestones. This is an attempt at milestone 1, to be followed by more. Some of the things in here might be already addressed in later commits.


src/droplet.F
```diff
@@ -1761,18 +1780,54 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind)
...
+         qf(k) = 0.0
+         Tpmean(k) = 0.0
+         radmean(k) = 0.0
+         qfsat(k) = 0.0
       end do
       !$acc end parallel
 
+
+      !! Number concentration
       !$acc parallel loop gang vector default(present) reduction(partnum)
```
johnmauff: It looks like a large number of variables Tf, qf, tpmean, radmean, qfsat now needs to be add to the $acc reduction clause.

src/droplet.F
```diff
@@ -1761,18 +1780,54 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind)
...
+         radmean(k) = radmean(k) + pdata(np,prrp)
+         if (pdata(np,prt) .gt. 200.0) then !Avoid this breaking if the temperature is unrealistic (like at initialization; other times this is a problem, but let it break elsewhere)
+         qfsat(k) = qfsat(k) + eslf(pdata(np,prprs),pdata(np,prt))*(mw/(ru*pdata(np,prt)*pdata(np,prrho)))
+         else
+         qfsat(k) = 0.0
+         end if
       end do
       !$acc end parallel
 
+      !! Temperature source
```
johnmauff, line 1808: This is a reduction that needs an openacc reduction clause.

src/droplet.F
```diff
@@ -1791,15 +1846,58 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind)
       if ( myid == 0 ) then
          call mpi_reduce(MPI_IN_PLACE,partnum,nk,MPI_REAL, &
```
johnmauff, line 1843: I would like to encourage the use of MPI_Allreduce, having a large number of duplicate mpi_reduce calls does not help code complexity.

src/droplet.F
```diff
@@ -571,18 +575,30 @@ subroutine droplet_driver(dt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,
...
 #ifdef MPI
-      call mpi_allreduce(nparcelsLocalActive,np_tmp,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
+   call mpi_allreduce(nparcelsLocalActive,np_tmp,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
 #endif
-      if (myid==0) then 
-         write(*,'(a8,3i)')  'DHR1:',myid,np_tmp,nparcelsLocalActive
-      end if
+
+   if (myid==0) then 
+   write(*,'(a8,3i)')  'DHR1:',myid,np_tmp,nparcelsLocalActive
```
sjsprecious: missing an `endif` here.

johnmauff (changes requested):

src/droplet.F
drichte2: Just a comment here: in subsequent stages of this overall pull request, all of the "1D" stats are combined into a single array called `droplet_diag_1`. So in a future sub-PR there will only be a single mpi_reduce call for an nnz-by-20 array, which contains all of them at once. Same idea as the `droplet_diag_0` array. So maybe just hold off on this comment until you see how it looks

johnmauff: Ok

src/writeout_nc.F
```diff
@@ -1643,15 +1643,17 @@ end subroutine writestat_nc
 !ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 !ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 
-      subroutine writedropdiag_nc(nrec,rtime,tnumpart,radavg,radsqr,radmin,radmax,Tpavg,Tpsqr,Tpmin,Tpmax,Tfavg,qfavg,prsavg,rhoavg,zhvec,numconc)
+      subroutine writedropdiag_nc(nrec,&
```
sjsprecious, line 1646: @drichte2 I thought you would eventually remove all the netCDF interface in CM1?

sjsprecious (changes requested):

src/writeout_nc.F
drichte2: I'm still not sure what to do about this. We had discussed this a while ago, but I just kept going with the netCDF since I'm not sure how else to output the drop_diag stats

src/droplet.F
drichte2: I don't get this -- my code definitely has this `endif`. I'm not sure how to change this...

sjsprecious: I mean we need two `end if` statements here. One to be paired with `if (myid==0) then` and the other to be paired with `if ( .not. use_wall_bc ) then`.

drichte2: Interesting. Well obviously that would have been repaired in one of the future PRs. Just commited a fix

src/droplet.F
sjsprecious: Thanks @johnmauff for catching this. I have added the OpenACC directives with reduction clause.

src/droplet.F
sjsprecious: Thanks @johnmauff for catching this. I have added them to the reduction clause.

sjsprecious:
With the additional changes for GPU directives, I am able to run this code on both CPU and GPU on Casper.

---

Stat output comparision

Identical variables: 63
four variables with largest error: PPMIN, PPMAX, DIVMIN, DIVMAX

Identical variables: 10

Identical variables: 512

---

The stat output comparison result is better than that in PR #101 but the 0th-order droplet diagnostic is worse. However, the error of two variables (`Tpavg` and `radmin`) is close to tolerance and I think it is acceptable.

johnmauff (approved):

sjsprecious (approved):

PR #104, opened 2023-04-19
Paper mods2: added load-imbalance timers

johnmauff:
This PR removes some dead code that was commented out and adds a new timer that measures load imbalance in the model.  It also eliminates a device-to-host copy that was unnecessary.  Note that to accurately measure load imbalance, a number of communication operators were grouped together  This also reduces the number of cpp  #ifdef _MPI in the code.

### DIFF
--- a/run/gust.node8.sh
+++ b/run/gust.node8.sh
@@ -7,17 +7,12 @@
 #PBS -A UNDM0006
 
 
-module purge
-module load nvhpc
-module load cray-mpich
-module load ncarcompilers
-module unload cuda
+module --force purge
+module load ncarenv/23.03
+module load nvhpc/23.1
+module load ncarcompilers/0.8.0
+module load cray-mpich/8.1.25
 
-export PALS_NRANKS=1024
-export PALS_PPN=128
-export PALS_DEPTH=1
-export PALS_CPU_BIND=depth
-
-#mpiexec ./cm1.exe --namelist namelist.figureE.cpu >& node=8.figureE.512x512.log
+mpiexec -n 1024 ./cpu.exe --namelist namelist.figureE.cpu >& gust.node8.figureE.512x512c.log
 #mpiexec ./cpu.exe --namelist namelist.figureD.cpu >& node=8.figureD.256.log
-mpiexec ./cm1.exe --namelist namelist.figureA.cpu >& gust.node8.figureA.256.log
+#mpiexec ./cm1.exe --namelist namelist.figureA.cpu >& gust.node8.figureA.256.log
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -4062,7 +4062,7 @@ end module cuda_rt
                     time_mpu2+time_mpv2+time_mpw2+time_mpp2+                &
                     time_mps1+time_mps3+time_mpq1+time_mptk1+                         &
                     time_mps2+time_mps4+time_mpq2+time_mptk2+time_mpb+                &
-                    time_dpc+time_droplet_reduce+time_parceli_reduce+                 &
+                    time_dpc+time_droplet_reduce+time_parceli_reduce+time_lb+         &
 #endif
                     time_advs+time_advu+time_advv+time_advw+                &
                     time_phys_H2D + time_phys_D2H
@@ -4387,6 +4387,10 @@ end module cuda_rt
         time_droplet = sum/float(numprocs)
 
         sum = 0.0
+        call MPI_REDUCE(time_lb ,sum,1,MPI_REAL,MPI_SUM,0,MPI_COMM_WORLD,ierr)
+        time_lb = sum/float(numprocs)
+
+        sum = 0.0
         call MPI_REDUCE(time_droplet_inject ,sum,1,MPI_REAL,MPI_SUM,0,MPI_COMM_WORLD,ierr)
         time_droplet_inject = sum/float(numprocs)
 
@@ -4473,7 +4477,7 @@ end module cuda_rt
                   time_mpu2+time_mpv2+time_mpw2+time_mpp2+                &
                   time_mps1+time_mps3+time_mpq1+time_mptk1+                     &
                   time_mps2+time_mps4+time_mpq2+time_mptk2+time_mpb+            &
-                  time_dpc+time_droplet_reduce+                                    &
+                  time_dpc+time_droplet_reduce+time_lb+                         &
 #endif
                   time_advs+time_advu+time_advv+time_advw+                &
                   time_phys_H2D+time_phys_D2H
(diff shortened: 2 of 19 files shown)


run/gust.node8.sh
```diff
@@ -7,17 +7,12 @@
...
-module purge
-module load nvhpc
-module load cray-mpich
-module load ncarcompilers
-module unload cuda
+module --force purge
+module load ncarenv/23.03
+module load nvhpc/23.1
+module load ncarcompilers/0.8.0
+module load cray-mpich/8.1.25
```
sjsprecious, line 14: For CPU run, shall we use the intel compiler instead for better performance?

src/cm1.F
```diff
@@ -4381,10 +4381,19 @@ program cm1
         sum = 0.0
         call MPI_REDUCE(time_parceli ,sum,1,MPI_REAL,MPI_SUM,0,MPI_COMM_WORLD,ierr)
         time_parceli = sum/float(numprocs)
+
         sum = 0.0
         call MPI_REDUCE(time_droplet ,sum,1,MPI_REAL,MPI_SUM,0,MPI_COMM_WORLD,ierr)
         time_droplet = sum/float(numprocs)
```
sjsprecious, line 4388: Does `time_droplet` include the timing of `time_lb` and `time_droplet_inject` below?

src/droplet.F
```diff
@@ -2167,6 +2177,7 @@ subroutine gauss_newton_2d(diffnorm,dt_nondim,taup_scale,rt_start,rt_zeros,flag,
                 v1(2) = rt_zeros(2)
 
         end do
+        !print *,'gauss_newton_2d: iterations: ',iterations
```
sjsprecious: remove this line

src/droplet.F
```diff
@@ -2482,7 +2493,7 @@ subroutine droplet_injection(pdata,xf,yf,dt,s10_xym,randomNumbers)
 
       my_reintro = floor(xrange*yrange*dt*totdrops/drop_mult)      !Number of droplets to be produced by this MPI task
 
-      my_reintro = 10
+      !JMD my_reintro = 10
```
sjsprecious: remove this line (also removed in David's PR)

src/solve3.F
```diff
@@ -405,13 +406,12 @@ subroutine solve3(nstep,rbufsz,num_soil_layers,                  &
       if(timestats.ge.1) time_prsrho=time_prsrho+mytime()
 
 
-        call bcs_GPU(rr)
+      call bcs_GPU(rr)
+      call bcs_GPU(rf)
```
sjsprecious, line 410: I saw that you moved many subroutine calls like this and I did not think they were answer changing. I am just curious about why you want to re-order these calling sequences and how you determine to move their locations up or down?

src/turb.F
```diff
@@ -2498,42 +2487,35 @@ subroutine tkekm(nstep,rtime,dt,ruh,rvh,rmh,zf,mf,rmf,zntmp,ust,ustt,rf,cmz,csz,
...
+!    if( nstep.eq.0 .and. myid.eq.0 )then
+!      !$acc update host (grdscl)
+!      print *
+!      print *,'  zf,grdscl:'
+!      i = 1
+!      j = 1
+!      do k=2,nk
+!        print *,k,(zf(i,j,k)-zf(i,j,1)),grdscl(i,j,k)
+!      enddo
+!    endif
```
sjsprecious: remove these lines

sjsprecious (commented):
Thanks @johnmauff for issuing this PR. I did not run a verification test but I did not think these changes were answer changing with my visual judgement. Just have a few clarification questions.

run/gust.node8.sh
johnmauff: Jian, we could you if want.  While on Cheyenne the Intel compiler offers significant advantages versus nvhpc, however for the AMD processors, I don't see much difference between Intel and the nvhpc compiler.

src/cm1.F
johnmauff: No, the time_droplet is independent of either time_lb (Time spent in load imbalance) or the injection time.  I wanted to see how expensive the droplet injection calculation really was.  We can certainly remove the separate time_drolet_inject once we are done optimizing the entire Droplet calculation.

src/droplet.F
johnmauff: Done

src/droplet.F
johnmauff: Done

src/turb.F
johnmauff: Done

src/solve3.F
johnmauff: The moving of subroutine calls around was associated with adding the load-imbalance timers, and the desire to reduce the number of #ifdef _MPI directives in the code.  So I grouped the packing routines bc*_GPU and the MPI boundary exchange calls together.

run/gust.node8.sh
sjsprecious: Oh really, I did not realize that. I am still under the impression that the intel compiler outperforms the nvhpc compiler significantly on the CPU but it looks like this is limited to the Intel's CPU only? This is important information to be recorded and tested in other models, too.

src/cm1.F
sjsprecious: Thanks for the clarification. Clear to me now.

src/solve3.F
sjsprecious: So does it mean that it does not matter if I call the `comm_*_GPU` subroutines before or after the `bc*_GPU` subroutines, as long as I group them together first?

run/gust.node8.sh
johnmauff: I was actually surprised as well.  It is illustrated in Figure 5 of the CM1 paper.  I have not checked for other codes yet.

src/solve3.F
johnmauff: Jian,  I kept the overall order between bc*_GPU and comm_*_GPU calls.  The order is critical, it just does not matter if all the operations on variable 'foo' are performed before the operations on variable 'bar'.  For example  

Old order 

call bcs_GPU(foo)
call comm_start_GPU(foo)
call bcs_GPU(bar)
call comm_start_GPU(bar)

New order:

call bcs_GPU(foo)
call bcs_GPU(bar)
call comm_start_GPU(foo)
call comm_start_GPU(bar)

sjsprecious: Thanks @johnmauff . That is clear to me now.

sjsprecious:

---

Stat output comparision

Identical variables: 63

Identical variables: 10

Identical variables: 512

---

The output is identical with that in PR #103.

sjsprecious (approved):

PR #105, opened 2023-04-24
David physics tmp

drichte2:
The next installment of the large PR. This one adds a new type of droplet statistic (droplet histograms) into the droplet_diag output.

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -111,7 +111,7 @@
       real :: xv,yv,zv,dV,wtx,wty,wtz,wtt
       real :: esl
 
-      real :: uval,vval,wval,qval,prsval,tval,rhoval
+      real :: uval,vval,wval,qval,prsval,tval,rhoval,rhval
       real :: sigdot
 
       logical, parameter :: debug = .false.
@@ -393,9 +393,11 @@
                                  xh,xf,yh,yf,zh,zf,zs,znt,sigma,sigmaf, &
                                  ua,va,wa,ta,qa,rho,prs,sigdot)
 
-!      if (rhval .lt. 0.5 .or. rhval .gt. 1.1) then
-!      write(*,'(a8,i,6e15.6)') 'DHR3',np,rhval,tval,rhoval,qval,mw,esa
-!      end if
+
+      rhval = (ru*tval*rhoval*qval)/(mw*eslf(prsval,tval))
+      if (rhval .lt. 0.5 .or. rhval .gt. 1.1) then
+      write(*,'(a8,i,8e15.6)') 'DHR3',np,rhval,tval,rhoval,qval,x3d,y3d,z3d,pdata(np,prrp)
+      end if
 
 
       call BE_integration(dt,np,pdata,pdata_locind,xf,yf,x3d,y3d,z3d,sig3d, &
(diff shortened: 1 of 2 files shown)


src/droplet.F
```diff
@@ -1645,6 +1644,12 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten)
       integer :: tnumpart,tnumdrop,tnumaerosol
       real :: radavg,radsqr,radmin,radmax,Tpavg,Tpmin,Tpmax,Tpsqr,Tfavg,qfavg,prsavg,rhoavg
       real :: partnum(nk),numconc(nk),Tpsrc(nk),qvsrc(nk),qf(nk),Tf(nk),radmean(nk),Tpmean(nk),qfsat(nk)
+
+      integer, parameter :: nbins = 514
```
johnmauff: I would like to see what the mean of nbins described in a comment.

johnmauff (approved):
@drichte2 This is exactly the size of PR that make review simple and fast.  Thank you for making it a manageable size.  Other than the need to add a comment about the meaning of nbins=514 I think this PR is ready.

src/droplet.F
```diff
@@ -1932,15 +1941,83 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten)
...
+
+      dr = (rmax10-rmin10)/(nbins-1)
+
+      bins_radius(1) = rmin10-dr
+      do i=1,nbins-1
+         bins_radius(i+1) = bins_radius(i) + dr
+      end do 
+
+      !Now that the bin centers are defined, loop through particles
+      do np=1,nparcelsLocalActive
```
sjsprecious, line 1966: either need an OpenACC directive for this loop or update`pdata` on the host first.

src/droplet.F
```diff
@@ -1932,15 +1941,83 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten)
...
+#endif
+
 #ifdef NETCDF
       ! JS: this subroutine is not updated with the new MPI
       !     interface for the droplets yet
 #ifdef MPI
       if ( myid == 0 ) then
 #endif
-      call writedropdiag_nc(nrec, &
+      call writedropdiag_nc(nrec,nbins, &
```
sjsprecious, line 2015: so the `hist_radius10` and `hist_radius` are always written to the netCDF file? In this case, we could not check its values & correctness through the log files.

sjsprecious (commented):
@drichte2 , thanks for this PR and it generally looks good to me, too. I have a quick question about how to evaluate / check the new droplet diagnostics.

src/droplet.F
drichte2: That's right, however the average droplet radius is already being compared, which is the mean of this distribution. In other words, if the average droplet radius is checking out, then that must mean that the histogram is also okay. We could always add a write statement within this routine if needed, but the bigger thing is to figure out where to put the droplet_diag in the future

sjsprecious: Thanks for your clarification. That is clear to me now.

sjsprecious (approved):

PR #106, opened 2023-04-24
Sub-PR 3

drichte2:
The main point of this is to move the droplet_diag write subroutine into droplets.F from writeout_nc.F. This allows for certain droplet diagnostics to be defined in the droplet module for convenience.


sjsprecious:

---

Stat output comparision

Identical variables: 62
variables with rel_diff > 1.e-6: 24

Identical variables: 10

Identical variables: 512

---

Compared to the PR #103, the stat output comparison result shows one fewer identical variable and one more variable with relative difference >1.e-6. The droplet diagnostic is consistent.

johnmauff:
Which diagnostic variable changed with this PR?

sjsprecious:
@johnmauff it is weird that I just re-ran the PR #103 on Casper and also got 62 identical variables and 24 variables with relative difference > 1.e-6 now. Unfortunately I did not save the log files for the "63 identical variables" case and thus I did not know which variable changed the result here.

sjsprecious:
@johnmauff, it turns out that the difference comes from the conda environment on Casper. If I run the verification test within a conda environment, there are "63 identical variables" between CPU and GPU runs. If I run the verification test without a conda environment,  there are "62 identical variables". The different stat variable is `MAXqv` and the relative difference is about 5e-6.

sjsprecious (approved):

johnmauff (approved):
This PR looks good.

PR #107, opened 2023-04-25
sub-PR 4

drichte2:
Added several more statistics, and created a single 1st order statistics array so that there is only a single mpi_reduce call in droplet_diag

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -1631,7 +1631,7 @@
 
       end subroutine project_feedback
 
-      subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten)
+      subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten,uten,vten,wten)
 
       use input
       use constants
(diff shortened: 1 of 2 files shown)


src/droplet.F
```diff
@@ -1830,20 +1868,32 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten)
...
+         m3src_tmp = 0.0
          Tpsrc_tmp = 0.0
          qvsrc_tmp = 0.0
          !$acc loop vector collapse(2) reduction(+:Tpsrc_tmp,qvsrc_tmp)
          do j=1,nj
-            do i=1,ni
-               Tpsrc_tmp = Tpsrc_tmp + dpten(i,j,k,1)
-               qvsrc_tmp = qvsrc_tmp + dpten(i,j,k,2)
-            end do
+         do i=1,ni
```
johnmauff: @drichte2 Could you add the necessary OpenACC directives here?

johnmauff (changes requested):

src/droplet.F
drichte2: Do you mean adding `m1src_tmp`, `m2src_tmp`, and `m3src_tmp` into the `reduction` call? I'm not sure what you mean otherwise

sjsprecious: I have just added them.

sjsprecious:

---

Stat output comparision

Identical variables: 62
variables with rel_diff <= 1.e-6: 1

Identical variables: 10
Two variables with largest error: Tpavg (~3.4e-6), radmin (~1.2e-5)

Identical variables: 512

---

Compared to the PR#106, the stat output comparison result shows one fewer identical variable and one more variable with relative difference <1.e-6 (variable name `TKEMAX`). The 0th-order droplet diagnostic shows the same variables with relative difference >1.e-6 but the magnitude here is larger. The 1st-order droplet diagnostic is consistent.

src/droplet.F
johnmauff: @drichte2, yes,the m1src_tmp, m2src_tmp, and m3src_tmp variables need to be added to the OpenACC directive just like the Tpsrc_tmp and qvsrc_tmp variables.  It looks like Jian beat you to it.

johnmauff (approved):

sjsprecious (approved):

PR #108, opened 2023-04-26
sub-PR 5

drichte2:
Hopefully there aren't too many things in this one. Several different miscellaneous items, including some new statistics.

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -16,7 +16,7 @@
   integer :: num100,num1000
   integer, parameter :: nbins = 514  !Bins for the droplet histograms
   real,dimension(nbins) :: bins_restime,hist_restime  !Histograms for droplet residence time
-  real,dimension(nbins) :: bins_radius,hist_radius,hist_radius10,hist_deathradius  !Histograms for droplet radius
+  real,dimension(nbins) :: bins_radius,hist_radius,hist_radius10,hist_fallrad  !Histograms for droplet radius
   real :: restime_min,restime_max,restime_min10,restime_max10,dhres !Bin parameters for droplet residence time
   real :: dhrad,rmin,rmax,rmin10,rmax10  !Bin parameters for droplet size
 
@@ -660,7 +660,7 @@
       real :: taup,Rep,Shp,diffnorm
       real :: rhop,volp,tp0,volp0
       real :: sig1
-      real :: top,bot,restime10_tmp
+      real :: top,bot,restime10_tmp,rad10_tmp
       integer :: i,j,k,ibin
       real :: taup_scale,dt_nondim,guess
       double precision :: rt_zeros(2),rt_start(2)
@@ -804,9 +804,9 @@
 
  
             !Log this dead particle into the relevant histograms
+            !First residence time
             restime10_tmp = log10(pdata(np,prtime))
 
-            !Figure out which bin it belongs to
             if (restime10_tmp .gt. restime_max10 + 0.5*dhres) then
                ibin = nbins
             elseif (restime10_tmp .lt. restime_min10 - 0.5*dhres) then
@@ -816,6 +816,20 @@
             end if
        
             hist_restime(ibin) = hist_restime(ibin) + pdata(np,prmult)
+
+
+            !The the droplet size at fallout
+            rad10_tmp = log10(pdata(np,prrp))
+        
+            if (rad10_tmp .gt. rmax10 + 0.5*dhrad) then
+               ibin = nbins
+            elseif (rad10_tmp .lt. rmin10 - 0.5*dhrad) then
+               ibin = 1
+            else
+               ibin = (floor((rad10_tmp-(rmin10-0.5*dhrad))/dhrad)+1)+1
+            end if
+
+            hist_fallrad(ibin) = hist_fallrad(ibin) + pdata(np,prmult)
 
          end if 
          pdata_locind(np,3) = undefined_index
@@ -2100,12 +2114,16 @@
                          MPI_SUM,0,MPI_COMM_WORLD,ierr)
           call mpi_reduce(MPI_IN_PLACE,hist_restime,nbins,MPI_REAL, &
                          MPI_SUM,0,MPI_COMM_WORLD,ierr)
+          call mpi_reduce(MPI_IN_PLACE,hist_fallrad,nbins,MPI_REAL, &
+                         MPI_SUM,0,MPI_COMM_WORLD,ierr)
        else
           call mpi_reduce(hist_radius,hist_radius,nbins,MPI_REAL, &
                          MPI_SUM,0,MPI_COMM_WORLD,ierr)
           call mpi_reduce(hist_radius10,hist_radius10,nbins,MPI_REAL, &
                          MPI_SUM,0,MPI_COMM_WORLD,ierr)
           call mpi_reduce(hist_restime,hist_restime,nbins,MPI_REAL, &
+                         MPI_SUM,0,MPI_COMM_WORLD,ierr)
+          call mpi_reduce(hist_fallrad,hist_fallrad,nbins,MPI_REAL, &
                          MPI_SUM,0,MPI_COMM_WORLD,ierr)
        end if
 #endif
@@ -2125,7 +2143,7 @@
 #endif
       !Now that it's done, remember to zero out the cumulative stats:
       hist_restime = 0.0
-      hist_deathradius = 0.0
+      hist_fallrad = 0.0
       num_destroyed = 0.0
       mass_destroyed = 0.0
       num_injected = 0.0
@@ -2954,7 +2972,7 @@
     !Initialize histogram to zero
     !(only those which accumulate between dropdiag I/O)
     hist_restime = 0.0
-    hist_deathradius = 0.0
+    hist_fallrad = 0.0
     num_destroyed = 0.0
     mass_destroyed = 0.0
     num_injected = 0.0
@@ -3233,6 +3251,10 @@
     call disp_err( nf90_put_att(ncid,varid,"long_name","total count for each residence time bin over the past statfrq time") , .true. )
     call disp_err( nf90_put_att(ncid,varid,"units","#") , .true. )
 
+    call disp_err( nf90_def_var(ncid,"hist_fallrad",nf90_float,(/binsid,timeid/),varid) , .true. )
+    call disp_err( nf90_put_att(ncid,varid,"long_name","total radius count of droplet which fell out over the past statfrq time") , .true. )
+    call disp_err( nf90_put_att(ncid,varid,"units","#") , .true. )
+
 
     call disp_err( nf90_put_att(ncid,NF90_GLOBAL,'CM1 version',trim(cm1version)) , .true. )
     call disp_err( nf90_enddef(ncid=ncid) , .true. )
@@ -3389,6 +3411,9 @@
 
     call disp_err( nf90_inq_varid(ncid,"hist_restime",varid) , .true. )
     call disp_err( nf90_put_var(ncid,varid,hist_restime,(/1,time_index/)) , .true. )
+
+    call disp_err( nf90_inq_varid(ncid,"hist_fallrad",varid) , .true. )
+    call disp_err( nf90_put_var(ncid,varid,hist_fallrad,(/1,time_index/)) , .true. )
 
     ! close file
 
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -406,9 +406,9 @@
 
 
       rhval = (ru*tval*rhoval*qval)/(mw*eslf(prsval,tval))
-      if (rhval .lt. 0.5 .or. rhval .gt. 1.1) then
-      write(*,'(a8,i,8e15.6)') 'DHR3',np,rhval,tval,rhoval,qval,x3d,y3d,z3d,pdata(np,prrp)
-      end if
+      !if (rhval .lt. 0.5 .or. rhval .gt. 1.1) then
+      !write(*,'(a8,i,8e15.6)') 'DHR3',np,rhval,tval,rhoval,qval,x3d,y3d,z3d,pdata(np,prrp)
+      !end if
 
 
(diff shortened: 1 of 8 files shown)


src/parcel.F
```diff
@@ -1221,8 +1221,6 @@ subroutine parcel_interp(dt,mtime,xh,uh,ruh,xf,uf,yh,vh,rvh,yf,vf, &
     nploop2:  &
     DO np=1,nparcelsLocal
 
-      pdata(np,prtime) = mtime
```
sjsprecious, line 1224: why removing this line?

src/param.F
```diff
@@ -6084,6 +6084,7 @@ subroutine param(namelist_input,                                  &
         prw = npvals
 
         npvals = npvals+1
+        npvars = npvars+1
```
sjsprecious, line 6087: why adding `npvars` by 1 here?

src/parcel.F
drichte2: I wanted each particle to know how long it have lived (i.e. stored in pdata). But then I realized that there was already a pdata "prtime" that wasn't being used, so I just repurposed it. It used to simply record the model time, but now I changed it so that it provides total accumulated time since the particle was created.

src/param.F
drichte2: See above, but it would need to be stored at restart. It's not a diagnostic (like the interpolated quantities are)

src/droplet.F
```diff
@@ -8,13 +8,20 @@ MODULE droplet_module
...
   integer :: num100,num1000
-  real :: u10_ssgf,num_destroyed,num_injected
+  integer, parameter :: nbins = 514  !Bins for the droplet histograms
+  real,dimension(nbins) :: bins_restime,hist_restime  !Histograms for droplet residence time
+  real,dimension(nbins) :: bins_radius,hist_radius,hist_radius10,hist_fallrad  !Histograms for droplet radius
+  real :: restime_min,restime_max,restime_min10,restime_max10,dhres !Bin parameters for droplet residence time
+  real :: dhrad,rmin,rmax,rmin10,rmax10  !Bin parameters for droplet size
+
+  real :: u10_ssgf 
+  real :: num_destroyed,num_injected,mass_destroyed,mass_injected !Cumulative particle counts between droplet_diag calls
```
sjsprecious, line 24: @drichte2 Do these variables have to be module variables? They are used in the `acc routine` and involved in the reduction operation, which makes the GPU implementation difficult. Would like to see @johnmauff's option, too.

src/param.F
sjsprecious: thanks for the clarification.

src/parcel.F
sjsprecious: thanks for the clarification.

src/droplet.F
drichte2: Technically, there would be a way to remove these as module variables, but it'd be a pain. They'd have to be sent out of the droplet_driver and then into droplet_diag, which would mean declaring them somewhere at the cm1.F level. I suppose this could be done, but this seemed like a more compact way of doing it. 

For what it's worth, @johnmauff said in an email when I asked about this, "I don't think this should cause a problem for the GPU port, however, I want to hear Jian's opinion as well.   We can just mark the new data structures as device resident and make sure that the correct version gets updated."

src/droplet.F
```diff
@@ -1661,14 +1695,14 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten,uten,vten,wten
...
-      real :: radavg,radsqr,radmin,radmax,Tpavg,Tpmin,Tpmax,Tpsqr,Tfavg,qfavg,prsavg,rhoavg,tnummult
-      real :: partnum(nk),numconc(nk),vp1mean(nk),vp2mean(nk),vp3mean(nk),vp1msqr(nk),vp2msqr(nk),vp3msqr(nk),Tpsrc(nk),qvsrc(nk),qf(nk),Tf(nk),radmean(nk),radmsqr(nk),Tpmean(nk),Tpmsqr(nk),qfsat(nk),m1src(nk),m2src(nk),m3src(nk)
+      integer :: tnumpart
+      real :: radavg,radsqr,radmin,radmax,Tpavg,Tpmin,Tpmax,Tpsqr,Tfavg,qfavg,prsavg,rhoavg,tnummult,tnumdrop,tnumaerosol
+      real :: partnum(nk),numconc(nk),vp1mean(nk),vp2mean(nk),vp3mean(nk),vp1msqr(nk),vp2msqr(nk),vp3msqr(nk),Tpsrc(nk),qvsrc(nk),qf(nk),Tf(nk),radmean(nk),radmsqr(nk),Tpmean(nk),Tpmsqr(nk),qfsat(nk),m1src(nk),m2src(nk),m3src(nk),qstarmean(nk)
 
       integer, parameter :: nbins = 514   !Number of bins to use for the droplet_diag histograms
-      integer, parameter :: num0 = 12, num1 = 19   !Number of 0th and 1st order statistics in droplet_diag
+      integer, parameter :: num0 = 18, num1 = 20   !Number of 0th and 1st order statistics in droplet_diag
       real,dimension(nbins) :: bins_radius,hist_radius,hist_radius10
```
sjsprecious: shall we remove these three variables here since they are module variables now?

drichte2: Yes, sorry -- that got missed when fixing the merge conflicts. Note that num0 and num1 need to stay where they are. Those that got moved to the module should be removed (in fact, I don't understand how this compiled for me...)

sjsprecious: Thanks @drichte2 for the clarification and fix.

src/droplet.F
sjsprecious: Generally I think John's comment is correct. However, in this particular case, those module variables are involved in a reduction operation in an acc routine, and I feel it is not trivial to get the code right on the GPU. I have added the necessary GPU directives (I think) and work on the testing now.

johnmauff: I am not particularly supportive of adding additional arguments that are passed through the subroutine argument list.  However, Jian is correct that the use of module variables complicates the GPU implementation.  Currently, CM1 has a qbudget array that stores a number of different global sums related to moisture.  Maybe something similar could be done with droplet statistics if Jian's testing on GPU proves problematic.

src/droplet.F
```diff
@@ -2083,48 +2146,51 @@ subroutine droplet_diag(rtime,nrec,zh,zf,pdata,pdata_locind,dpten,uten,vten,wten
          rad10_tmp = log10(pdata(np,prrp))
         
          !Figure out which bin it belongs to
-         if (rad10_tmp .gt. rmax10 + 0.5*dr) then
+         if (rad10_tmp .gt. rmax10 + 0.5*dhrad) then
```
sjsprecious, line 2127: @drichte2 Shall we remove line 2128 to 2139 here? I saw similar calculations in the `initialize_droplet_histograms` subroutine and they did not seem to vary with different time steps.

drichte2: This is strange... Yes, you are absolutely right. But somehow these lines aren't in my local version that I had pushed. I must have messed up something when I was merging? I'd change this myself, but I don't see how. Could have this come back in when you merged gpu-opt back into this branch?

sjsprecious: Thanks @drichte2. I did not know why this happened, either. I just removed these lines and pushed the changes to your branch. Could you pull the latest version of your branch to your local machine and check if everything looks good to you now?

drichte2: Just pulled, looks good to me!

sjsprecious: Thanks!

sjsprecious:

---

Stat output comparison

Identical variables: 63

Identical variables: 9
variable with rel_diff > 1.e-6: 3
three variables with largest error: Tpavg (~3.4e-6), radmin (~6.2e-6), prsavg (~1.1e-6)

Identical variables: 511
variable with rel_diff > 1.e-6: 1
one variables with largest error: numconc(1) (~1.0)

---

Compared to the PR #109, the stat output comparison result is consistent. The 0th-order droplet diagnostic shows one fewer identical variable and one more variable with relative difference >1.e-6, which is `prsavg`. The 1st-order droplet diagnostic shows one inconsistent variable, which is `numconc(1)`. The relative difference is close to 100%. The value of `numconc(1)` in the CPU run is **0.150681E-04** while that in the GPU run is **0.152588E-07**. I am not sure whether this difference is caused by the mismatch of `prsavg` variable or not. I am also not sure whether the difference of the droplet diagnostic here is even expected or not, though the `init3d` subroutine is tweaked slightly here. Need to see @drichte2 and @johnmauff's opinions on these results.

drichte2:
@sjsprecious I'm pretty sure the answer is yes, but `tnumpart` is part of the 0th order diagnostics, right? And that comes out exactly the same (it'd better)? The difference in `prsavg` should have nothing to do with the 1st order diagnostic. 

It would be very weird if `tnumpart` was the same but only one element of `numconc` was off, since the sum of the `numconc` array should be equal to `tnumpart` divided by the domain volume.

drichte2:
@sjsprecious For this issue, maybe instead of checking `numconc` you can replace it with `partnum`? This variable simply says how many particles are within each height range. The sum of this array should give you exactly `tnumpart`, and it's `partnum` which is subject to reduction operations (and therefore where the difference might come from)

sjsprecious:
> @sjsprecious I'm pretty sure the answer is yes, but `tnumpart` is part of the 0th order diagnostics, right? And that comes out exactly the same (it'd better)? The difference in `prsavg` should have nothing to do with the 1st order diagnostic.
> 
> It would be very weird if `tnumpart` was the same but only one element of `numconc` was off, since the sum of the `numconc` array should be equal to `tnumpart` divided by the domain volume.

Hi @drichte2, thanks for your quick reply. Yes, `tnumpart` matches exactly between CPU and GPU runs, which makes me think if something is wrong when calculating the domain volume. I originally thought `prsavg` was the culprit but it seems not now.

sjsprecious:
> @sjsprecious For this issue, maybe instead of checking `numconc` you can replace it with `partnum`? This variable simply says how many particles are within each height range. The sum of this array should give you exactly `tnumpart`, and it's `partnum` which is subject to reduction operations (and therefore where the difference might come from)

Hi @drichte2, thanks for your suggestion. I just realized that I commented out some GPU directives in the `droplet_diag` subroutine and therefore the 1st-order droplet diagnostic was problematic. Somehow I could not get those directives work for the changes in this PR and I would need more time to debug them. Sorry for the confusion.

sjsprecious:
Hmm, it turns out that the GPU code with the changes here no longer compiles on Casper (nvhpc/22.2 + cuda/11.6), but still compiles on Gust (nvhpc/23.1 + cuda/11.7). Thus maybe our GPU changes here trigger a bug in the old nvhpc compiler.

By using the namelist `namelist.restC` and without the conda environment, the verification result on **Gust** looks below:

---

CPU (intel, 2 MPI ranks) vs. GPU (nvhpc, 2 MPI ranks + 2 A100 GPUs)

Stat output comparison

Identical variables: 59
variables with rel_diff > 1.e-6: 22
variables with rel_diff <= 1.e-6: 5

Identical variables: 8
variable with rel_diff > 1.e-6: 3
variable with rel_diff <= 1.e-6: 1
three variables with largest error: Tpavg (~3.4e-6), radmin (~6.2e-6), prsavg (~1.1e-6)

Identical variables: 512

---

Compared to the PR https://github.com/george-bryan/CM1/pull/109, the stat output comparison result shows four fewer identical variable, one fewer variable with relative difference >1.e-6, and five more variables with relative difference <1.e-6. These five variables are `THPMAX`, `STHPMX`, `TKEMAX` , `RHMIN` and `THPMIN`. The 0th-order droplet diagnostic shows two fewer identical variables, one more variable with relative difference >1.e-6 (which is `prsavg`), and one more variable with relative difference <1.e-6 (which is `radavg`). The 1st-order droplet diagnostic is consistent. Since I switch to a different machine, I expect that the verification results may change and most non-identical variables are still within the tolerance. Still need to see @drichte2 and @johnmauff's comments on these results.

drichte2:
> droplet diagnostic shows two fewer identical variables, one more variable with relative difference >1.e-6 (which is `prsavg`), and one more variable with relative difference <1.e-6 (which is `radavg`). The 1st-order droplet diagnostic is consistent. Since I switch to a different machine, I expect that the verification results may change and most non-identical variables are still within the tolerance. Still need to see @drichte2 and @johnmauff's comments on these results.

@sjsprecious Just to be clear, these are relative differences, right? For example the droplet radii are O(1e-6) so an absolute difference that large is like 100% relative error. 

Also, just to confirm: these are a comparison between CPU and GPU of this most recent PR -- not a comparison of the previous version to the PR?

Assuming these are true, none of those numbers are particularly concerning to me. It seems that those that are larger than 1e-6 are still O(1e-6).

sjsprecious:
> > droplet diagnostic shows two fewer identical variables, one more variable with relative difference >1.e-6 (which is `prsavg`), and one more variable with relative difference <1.e-6 (which is `radavg`). The 1st-order droplet diagnostic is consistent. Since I switch to a different machine, I expect that the verification results may change and most non-identical variables are still within the tolerance. Still need to see @drichte2 and @johnmauff's comments on these results.
> 
> @sjsprecious Just to be clear, these are relative differences, right? For example the droplet radii are O(1e-6) so an absolute difference that large is like 100% relative error.
> 
> Also, just to confirm: these are a comparison between CPU and GPU of this most recent PR -- not a comparison of the previous version to the PR?
> 
> Assuming these are true, none of those numbers are particularly concerning to me. It seems that those that are larger than 1e-6 are still O(1e-6).

@drichte2 , yes, all the difference of statistics here is relative difference.

And yes, the comparison here is between CPU and GPU runs for this PR.

Glad to know that these numbers look fine to you. Will wait for John's input before I approve this PR.

sjsprecious (approved):

johnmauff (approved):
Based on Jian extensive review and testing, and discussion with him, I approve this PR.

PR #109, opened 2023-04-26
Update droplet I/O and droplet diagnostic for large number of droplets

sjsprecious:
This PR fixes the issue of crashed run when there are ~36M droplets over the whole domain.

The new droplet I/O interface in the `parcel.F` code is updated based on the version used in the `restart.F` code.

In addition, the GPU code of droplet diagnostic is also revised to use the `acc atomic update` to avoid blowing up the GPU memory.

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -31,6 +31,8 @@
   !For testing: set this parameter to ".false." to make droplet injection deterministic (not random)
   logical, parameter :: random_droplets = .true.
 
+  !$acc declare create(num100,num1000)
+
   CONTAINS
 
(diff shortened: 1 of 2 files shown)


sjsprecious:

---

Stat output comparison

Identical variables: 63

Identical variables: 10

Identical variables: 512

---

Compared to the PR #107, the stat output comparison result shows one more identical variable. The 0th-order droplet diagnostic shows the same variables with relative difference >1.e-6 but the magnitude here is now smaller. The 1st-order droplet diagnostic is consistent.

johnmauff (approved):
Thanks for fixing the I/O issues with large droplet counts.

sherimickelson (approved):
Thanks, @sjsprecious for fixing this quickly!

PR #111, opened 2023-05-05
Namelist for ASD

drichte2:
Provide namelist and input_sounding for ASD. Also move some variables from code to namelist


src/Makefile
```diff
@@ -156,7 +156,7 @@ endif
 ifeq (${FC},ifort)
     # replace -xHost with -march=core-avx2 to make sure the same code builds on
     # both Intel and AMD CPUs 
-    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
```
johnmauff: Could we revert this back to the -march=core-avx2 option?  The current setting will not work for AMD processors.

johnmauff (changes requested):

src/Makefile
drichte2: Arg yes, I keep forgetting this before pushing

sjsprecious:
@drichte2 , shall we revise the source code somewhere to read in the `input_sounding_ASD` namelist rather than the default `input_sounding` namelist?

drichte2:
> @drichte2 , shall we revise the source code somewhere to read in the `input_sounding_ASD` namelist rather than the default `input_sounding` namelist?

Sorry @sjsprecious I somehow had missed this. I don't think this is necessary. My plan for running the ASD cases is to have a separate directory for each run, and each directory having its version of the file named `input_sounding'. That way the same executable can be run on all of them in their local run directories

sjsprecious:
@drichte2 , thanks for your reply. That sounds good to me.

drichte2:
@sjsprecious just a heads up: I've been doing more tests today, and pulled your merge with gpu-opt from 2 hours ago -- but now my injection case is no longer working. I noticed that this had `use_wall_bc=true`, but when I turned that off, I'm getting a crash now just after the first droplet injection for the `namelist_ASD.input`

sjsprecious:
@drichte2 thanks for your update. I just merged John's PR #113 into the `gpu-opt` branch and it may conflict with your code here. I am working on the revision of your `david_physics_tmp` branch to address those issues and will let you know once it works again.

sjsprecious:
Hi, @drichte2 , I pushed some changes to your `david_physics_tmp` branch and it worked now for the ASD namelist. However, the droplet injection subroutine does not seem to inject any droplets in this case. Could you please pull the latest version of `david_physics_tmp` branch to your local machine and test it? Thanks.

drichte2:
> Hi, @drichte2 , I pushed some changes to your `david_physics_tmp` branch and it worked now for the ASD namelist. However, the droplet injection subroutine does not seem to inject any droplets in this case. Could you please pull the latest version of `david_physics_tmp` branch to your local machine and test it? Thanks.

@sjsprecious I just checked it and it still crashes as soon as the particle injection starts up at t = 1800s. It's throwing a floating point overflow error at one of the basic droplet calculations. I can look more into it tomorrow, but this wasn't happening before the changes introduced by @johnmauff 's PR

johnmauff (approved):
I would like to move forward based on testing by @drichte2

johnmauff:
@sjsprecious, I think we should accept this PR based as is and address any additional issues as they arise.

sjsprecious:
Thanks @johnmauff . I will approve this PR and merge it later.

PR #116, opened 2023-06-09
Drop inject

johnmauff:
This PR GPU enables the droplet injection code.  Correctness is measured using the namelist_ASD.verify namelist on Gust using either 4 CPU cores of 4 GPU devices.  The verification statistics are provided below.  Note that comparisions are provided both using two different compilers (Intel & NVHPC) and using the NVHPC compiler on GPU and CPU.  These comparisions demonstrate that any differences between CPU and GPU is similar to the differences observed using different compilers.  

nvhpc versus ifort on CPU

Metrics
37 stat variables are identical.

0th-order
6 stat variables show a mean relative difference > 1e-06

1st-order

CPU versus GPU using nvhpc

Metrics

0th-order
2 stat variables show a mean relative difference <= 1e-06

1st-order

### DIFF
--- a/run/namelist_ASD.verify
+++ b/run/namelist_ASD.verify
@@ -1,0 +1,369 @@
+
+ &param0
+ nx           =    128,
+ ny           =    128,
+ nz           =    128,
+ ppnode       =      64,
+ timeformat   =       1,
+ timestats    =       1,
+ terrain_flag = .false.,
+ procfiles    = .false.,
+ /
+
+ &param1
+ dx     =    40.0,
+ dy     =    40.0,
+ dz     =    20.0,
+ dtl    =   0.5,
+ timax  = 600.0,
+ run_time =  -999.9,
+ tapfrq =   1000.0,
+ rstfrq = 10800.0,
+ statfrq =  60.0,
+ prclfrq =   60.0,
+ /
+
+ &param2
+ cm1setup  =  1,
+ testcase  =  10,
+ adapt_dt  =  0,
+ irst      =  0,
+ rstnum    =  1,
+ iconly    =  0,
+ hadvordrs =  5,
+ vadvordrs =  5,
+ hadvordrv =  5,
+ vadvordrv =  5,
+ advwenos  =  2,
+ advwenov  =  0,
+ weno_order = 5,
+ apmasscon =  0,
+ idiff     =  0,
+ mdiff     =  0,
+ difforder =  6,
+ imoist    =  1,
+ ipbl      =  0,
+ sgsmodel  =  1,
+ tconfig   =  1,
+ bcturbs   =  1,
+ horizturb =  0,
+ doimpl    =  0,
+ irdamp    =  1,
+ hrdamp    =  0,
+ psolver   =  2,
+ ptype     =  0,
+ ihail     =  0,
+ iautoc    =  0,
+ icor      =  1,
+ lspgrad   =  3,
+ eqtset    =  1,
+ idiss     =  0,
+ efall     =  0,
+ rterm     =  0,
+ wbc       =  1,
+ ebc       =  1,
+ sbc       =  1,
+ nbc       =  1,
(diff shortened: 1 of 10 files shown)


src/droplet.F
```diff
@@ -3109,16 +3116,24 @@ subroutine droplet_injection(pdata,xf,yf,dt,time_since_inject,randomNumbers,mtim
...
-         do np=nparcelsLocalActive+1,nparcelsLocalActive+my_reintro
+         !$acc data copyin(dFssum,binsdata)
+         !$acc parallel loop gang vector default(present) reduction(+:mass_injected)
+         do j=1,my_reintro
+            np=nparcelsLocalActive+j
             if (random_droplets) then
+               ! initialize a new random number generator for each
+               ! injected droplet
+               local_seed = (1000 * int(mtime)) + (10000*myid) + j
+               randomNumbers = new_RandomNumberSequence(seed = local_seed)
```
sjsprecious: Not sure if we want to shuffle the seeds every time we call the `droplet_injection` subroutine. Does it make a difference here?

johnmauff: My motivation is to make sure we have a deterministic sequence of random numbers.  If we do it only once before the loop, then there will potentially be duplicate droplets on each set of threads, right?

sjsprecious: Thanks @johnmauff . After a second thought, I think you are right that without setting a unique seed for each CUDA thread, the result may not be deterministic because the sequence of pulling random number by each CUDA thread is not guaranteed.

src/misclibs.F
```diff
@@ -3037,10 +3037,13 @@ subroutine calccfl_fold(nstat,rstat,dt,acfl,uh,vh,mh,ua,va,wa,writeit)
...
-      imaxt = 0
-      jmaxt = 0
-      kmaxt = 0
+      !$acc parallel loop gang vector default(present)
+      do index=1,ni*nj
+        cfl(index) = -99999.0
+        imaxt(index) = 0
+        jmaxt(index) = 0
+        kmaxt(index) = 0
+      enddo
```
sjsprecious, line 3046: According to @johnmauff and @george-bryan , these calculations were not done correctly on GPU before and led to a CFL violation error later. The changes here fix this CFL error.

run/namelist_ASD.verify
```diff
@@ -0,0 +1,369 @@
+
+ &param0
+ nx           =    128,
+ ny           =    128,
+ nz           =    128,
+ ppnode       =      64,
```
sjsprecious: Shall we set `ppnode = 4`?

src/droplet.F
```diff
@@ -356,7 +359,7 @@ subroutine droplet_driver(dt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,
 #else
     !JMD WARNING: Loop does not yet match CPU version.
     !$acc parallel default(present)
-    !$acc loop gang vector reduction(+:num_fallout) &
+    !$acc loop gang vector reduction(+:num_fallout) reduction(max:errcode) &
```
sjsprecious, line 362: Do we need `errcode` besides debugging purpose?

johnmauff: @sjsprecious, I think we do need to keep errcode variable.  Currently, if a bug occurs in the section, it is nearly impossible to locate the source of the bug due to the high volume of output generated from each CUDA thread.

run/namelist_ASD.verify
johnmauff: Sure, I just updated the namelist file.

src/droplet.F
sjsprecious: Gotcha! Thanks for the clarification.

sjsprecious (approved):
Thanks @johnmauff for issuing this PR. It looks good to me and I could reproduce the verification results in the description section.

PR #118, opened 2023-06-17
Fix a bug in the droplet initialization

sjsprecious:
As reported by @johnmauff , the CPU/GPU code crashed when trying to initialize more than 100 millions droplets with the `namelist_ASD.verify` namelist.

The issue is that some droplets may fall exactly on the eastmost or northmost boundary of the whole domain, which won't be owned by any MPI rank based on the current implementation.

The changes in this PR fix this issue and now I could initialize 160 millions droplets. However, the code still crashes for other bugs.

### DIFF
--- a/src/init3d.F
+++ b/src/init3d.F
@@ -427,10 +427,19 @@
           rand = getRandomReal(randomNumbers)
           tmpz = rand*injectHMax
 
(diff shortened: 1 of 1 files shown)


johnmauff (approved):
Have not checked, but if it passes George's error checks I am fine with it.

PR #121, opened 2023-06-23
Duplicate code elimination in bc.F and comm.F

johnmauff:
This PR eliminates duplicate routines in the comm.F and bc.F modules.  It does this by enabling the runtime activation/deactivation of OpenACC directives via the if conditionals provided by the OpenACC standard.

The verification statistics match those from PR #122. 

ifort CPU versus nvhpc GPU
===========================

Metrics
-------
36 stat variables are identical.

0th-order
---------
2 stat variables show a mean relative difference <= 1e-06

1st-order
---------

### DIFF
--- a/src/base.F
+++ b/src/base.F
@@ -34,8 +34,8 @@
                       sw31,sw32,se31,se32,ss31,ss32,sn31,sn32)
       use input
       use constants
-      use bc_module, only: bcu_GPU, bcv_GPU, bcs_GPU, extrapbcs, &
-          bcu2_GPU, bcv2_GPU
+      use bc_module, only: bcu, bcv, bcs, extrapbcs, &
+          bcu2, bcv2
       use comm_module
       use goddard_module, only : T0K,T00K,RT0
       use cm1libs , only : rslf,rsif
@@ -2028,13 +2028,13 @@
 !  fill in ghost cells
 
       !$acc data copy(pi0,prs0,th0,qv0,qc0,qi0,rh0)
-      call bcs_GPU(pi0)
-      call bcs_GPU(prs0)
-      call bcs_GPU(th0)
-      call bcs_GPU(qv0)
-      call bcs_GPU(qc0)
-      call bcs_GPU(qi0)
-      call bcs_GPU(rh0)
+      call bcs(pi0)
+      call bcs(prs0)
+      call bcs(th0)
+      call bcs(qv0)
+      call bcs(qc0)
+      call bcs(qi0)
+      call bcs(rh0)
       !$acc end data
        
 
@@ -2129,7 +2129,7 @@
     enddo
 
       !$acc data copy(rho0)
-      call bcs_GPU(rho0)
+      call bcs(rho0)
       !$acc end data
 #ifdef MPI
       call comm_all_s(rho0,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,  &
@@ -2660,8 +2660,8 @@
 !  Fill in ghost cells
 
       !$acc data copy(u0,v0)
-      call bcu_GPU(u0)
-      call bcv_GPU(v0)
+      call bcu(u0)
+      call bcv(v0)
       !$acc end data
 
       !--------
@@ -2673,7 +2673,7 @@
       call getcorneru3(u0,n3w1(1,1,1),n3w2(1,1,1),n3e1(1,1,1),n3e2(1,1,1),  &
                           s3w1(1,1,1),s3w2(1,1,1),s3e1(1,1,1),s3e2(1,1,1))
       !$acc data copy(u0)
-      call bcu2_GPU(u0)
+      call bcu2(u0)
       !$acc end data
 #endif
 !$omp parallel do default(shared)   &
@@ -2693,7 +2693,7 @@
       call getcornerv3(v0,n3w1(1,1,1),n3w2(1,1,1),n3e1(1,1,1),n3e2(1,1,1),  &
                           s3w1(1,1,1),s3w2(1,1,1),s3e1(1,1,1),s3e2(1,1,1))
       !$acc data copy(v0)
-      call bcv2_GPU(v0)
+      call bcv2(v0)
       !$acc end data
 #endif
 !$omp parallel do default(shared)   &
--- a/src/base.F
+++ b/src/base.F
@@ -36,7 +36,7 @@
       use constants
       use bc_module, only: bcu, bcv, bcs, extrapbcs, &
           bcu2, bcv2
-      use comm_module, only: comm_all_s_GPU, getcorneru3,getcornerv3, &
+      use comm_module, only: comm_all_s_GPU,getcorneru3_GPU,getcornerv3_GPU, &
           comm_3u_start,comm_3u_end,comm_3v_start,comm_3v_end
       use goddard_module, only : T0K,T00K,RT0
       use cm1libs , only : rslf,rsif
@@ -2671,8 +2671,8 @@
                             us31,us32,un31,un32,reqs_u)
       call comm_3u_end(u0,uw31,uw32,ue31,ue32,   &
                           us31,us32,un31,un32,reqs_u)
-      call getcorneru3(u0,n3w1(1,1,1),n3w2(1,1,1),n3e1(1,1,1),n3e2(1,1,1),  &
-                          s3w1(1,1,1),s3w2(1,1,1),s3e1(1,1,1),s3e2(1,1,1))
+      call getcorneru3_GPU(u0,n3w1(1,1,1),n3w2(1,1,1),n3e1(1,1,1),n3e2(1,1,1),  &
+                          s3w1(1,1,1),s3w2(1,1,1),s3e1(1,1,1),s3e2(1,1,1),device=.false.)
       !$acc data copy(u0)
       call bcu2(u0)
       !$acc end data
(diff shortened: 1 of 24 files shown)


johnmauff:
@sjsprecious  I have verified that this PR has the same correctness characteristics as PR #122.  Can you take a look?  Note that there will be additional PRs like this one that involve cleanup in the comm.F file

src/base.F
```diff
@@ -2027,47 +2028,43 @@ subroutine base(zh,mh,rmh,c1,c2,zf,mf,rho0s,pi0s,prs0s,rth0s,         &
...
-      call bcs_GPU(qi0)
-      call bcs_GPU(rh0)
-      !$acc end data
+      call bcs(pi0, device=.false.)
+      call bcs(prs0,device=.false.)
+      call bcs(th0, device=.false.)
+      call bcs(qv0, device=.false.)
+      call bcs(qc0, device=.false.)
+      call bcs(qi0, device=.false.)
+      call bcs(rh0, device=.false.)
```
sjsprecious, line 2037: If in the future we want to run `bcs` on the GPU, do we need to manually copy those variables first as in the original code? I have this question for most changes of this PR.

src/hifrq.F
```diff
@@ -934,10 +934,10 @@ subroutine writeout_hifrq(                                                 &
 
       !c-c-c-c-c-c-c-c-c-c
 
-      call bcs(sadv)
+      call bcs(sadv,device=.false.)
 #ifdef MPI
-      call comm_3s_start(sadv,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,reqs_s)
-      call comm_3s_end(  sadv,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,reqs_s)
+      call comm_3s_start_GPU(sadv,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,reqs_s,device=.false.)
```
sjsprecious, line 939: It looks confusing to me that we are using a **XXX_GPU** subroutine but setting `device=.false.`.

src/param.F
```diff
@@ -8279,12 +8279,12 @@ subroutine param(namelist_input,                                  &
       enddo
       enddo
       enddo
-      !$acc data copy(cc2,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,n3w1,n3w2,n3e1,n3e2,s3w1,s3w2,s3e1,s3e2,reqs_s)
-      call bcs_GPU(cc2)
-      !$acc end data
+      !!$acc data copy(cc2,sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,n3w1,n3w2,n3e1,n3e2,s3w1,s3w2,s3e1,s3e2,reqs_s)
```
sjsprecious, line 8282: remove unused directive

src/param.F
```diff
@@ -8279,12 +8279,12 @@ subroutine param(namelist_input,                                  &
...
+      call bcs(cc2,device=.false.)
+      !!$acc end data
```
sjsprecious, line 8284: remove unused directive

src/sounde.F
```diff
@@ -982,7 +982,7 @@ subroutine sounde(dt,xh,arh1,arh2,uh,ruh,xf,uf,yh,vh,rvh,yf,vf,     &
 
         IF( n.lt.nloop )THEN
           if(timestats.ge.1) time_sound=time_sound+mytime()
-!!!          call bcs(ppd)
+!!!          call bcs_GPU(ppd,device=.false.)
```
sjsprecious, line 985: shall we just remove this subroutine call?

src/turb.F
```diff
@@ -2837,10 +2837,8 @@ subroutine turbsmag(nstep,rtime,dt,dosfcflx,ruh,rvh,rmh,mf,rmf,th0,thflux,qvflux
 !--------------------------------------------------------------
 
       if(timestats.ge.1) time_turb=time_turb+mytime()
-      !$acc data copy(kmh,kmv)
-      call bcw_GPU(kmh,1)
-      call bcw_GPU(kmv,1)
-      !$acc end data
+      call bcw(kmh,1)
+      call bcw(kmv,1)
```
sjsprecious, line 2841: In the original code, `kmh` and `kmv` need to be copied between CPU and GPU manually. Is it done somewhere else in your PR?

sjsprecious: or are we missing the `device=.false.` argument here?

src/turbnba.F
```diff
@@ -230,8 +226,8 @@ subroutine turbnba(nstep,uh,ruh,uf,ruf,vh,rvh,vf,rvf,mh,rmh,mf,rmf,zf,c1,c2,rho,
       enddo
 
 #ifdef MPI
+      call bct2(tkea,device=.false.)
       !$acc data copy(tkea)
-      call bct2_GPU(tkea)
       call getcornert_GPU(tkea,nw1,nw2,ne1,ne2,sw1,sw2,se1,se2)
```
sjsprecious, line 231: Can we modify the `getcornert_GPU` subroutine as other **bcXXX** subroutine to take the `device` argument and simplify the code here?

src/bc.F
```diff
@@ -26,23 +19,25 @@ MODULE bc_module
...
       implicit none
 
       real, dimension(ib:ie,jb:je,kb:ke) :: p
+      logical, optional, intent(in) :: device
 
       integer i,j,k
-      !$acc declare present(p)
+      logical :: openacc 
 
+      call SetGPUParams(openacc,device)
```
sjsprecious, line 33: Maybe we need to add `!$acc declare copyin(XXX)` for all the **bcXX** subroutines here?

src/bc.F
```diff
@@ -3656,10 +2947,26 @@ subroutine bcs2_GPU(s)
...
+         logical, optional :: device
+
+      if(present(device)) then
+         if(device) then
+            openacc = .true.
+         else
+            openacc = .false.
+         endif
+      else
+         openacc = .true.
```
sjsprecious, line 2966: This is just a preference question: is `.false.` or `.true.` better as the default value here?

sjsprecious (commented):
Thanks @johnmauff for working on this huge PR and cleaning up the code significantly. I could confirm that the verification result is consistent with the latest `gpu-opt` branch.

However, I have a few clarification questions before I approve this PR.

src/hifrq.F
sjsprecious: Can we call **XXX** subroutine and set `device=.true.` instead?

src/base.F
johnmauff: The use of the device keyword allows us to call a single version of bcs() on either GPU or CPU resident data.   The subroutines in the base.F are called during initialization, hence why device=.false.

src/hifrq.F
johnmauff: I understand the confusion as this is a code that is in transition.  Right now I am replacing all comm_3s_start calls with new versions that are currently named comm_3s_start_GPU.   I am using the device keyword to indicate if the variables are GPU or CPU resident.  Once I replace all duplicate subroutines in the comm.F file with the flexible variants, I will eliminate the _GPU naming structure.

src/param.F
johnmauff: I have made this change in a more recent set of commits, that is not currently included in this PR.

src/param.F
johnmauff: I have made this change in a more recent set of commits, that is not currently include in this PR.

src/sounde.F
johnmauff: Sure, I can clean this up in the next commit.

src/turb.F
johnmauff: I found these inconsistencies as well.  Currently, the code is never called which is why this does not matter.  We probably need to mark this code as not yet supported on the GPU.

src/turbnba.F
johnmauff: This will be done in a subsequent PR.  I just have not gotten to upgrading all subroutines in comm.F yet.

src/bc.F
johnmauff: I don't quite understand this comment.  Maybe we need to discuss on a video call.

src/bc.F
johnmauff: This is probably a longer discussion that we can have in person versus through github.

src/base.F
sjsprecious: Thanks for the clarification.

src/hifrq.F
sjsprecious: Thanks for the clarification.

src/param.F
sjsprecious: Thanks for the clarification.

src/param.F
sjsprecious: Thanks for the clarification.

src/turb.F
sjsprecious: Thanks for the clarification. It is better to mark it now before we forgot.

src/turbnba.F
sjsprecious: Thanks for the clarification.

src/bc.F
sjsprecious: After a discussion with @johnmauff , we do not need the `!$acc declare` statement in the **bcXX** subroutines and all the data copy should happen in the `cm1.F` code.

The `default (present)` clause checks that all the variables for the **bcXX** subroutines should be GPU-resident and an error message will be issued if it is not the case.

src/bc.F
sjsprecious: After the discussion with @johnmauff , it seems reasonable to set `.true.` as the default value and let the user specify which subroutines they want to run on CPU instead.

sjsprecious (approved):

PR #122, opened 2023-06-23
Overflow fixes from George

drichte2:
This has multiple additions from George:
- A conversion of a number of variables to double precision and additional checks to prevent overflow issues in droplet iteration
- Computing velocity tendencies at the node center (different than humidity and temperature tendencies) and adding them to the velocity solvers
- Rearranging a few things for making droplet stats more convenient

The correctness statistics for this PR is as follows:

ifort on CPU versus nvhpc on CPU
==========================

Metrics
--------

0th-order
----------
5 stat variables show a mean relative difference > 1e-06
4 stat variables show a mean relative difference <= 1e-06

1st-order
---------

nvhpc on CPU versus nvhpc on GPU
============================

Metrics
--------

0th-order
----------

1st-order
---------

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -55,7 +55,7 @@
                                nw1,nw2,ne1,ne2,sw1,sw2,se1,se2,reqs_p,                    &
                                sw31,sw32,se31,se32,ss31,ss32,sn31,sn32,                   &
                                n3w1,n3w2,n3e1,n3e2,s3w1,s3w2,s3e1,s3e2,reqs_s,            &
-                               uten,vten,wten,thten,qten,dten,pdata_locind)
+                               dpten,pdata_locind)
       use input, only : ib,ie,jb,je,kb,ke,ibl,iel,jbl,jel,numq,ni,nj,nk,imp,jmp,kmp,rmp,cmp, &
           kmt,npvals,npvars,cgs1,cgs2,cgs3,cgt1,cgt2,cgt3,bbc,tbc,imove,zt,rzt, &
           umove,vmove,nqv,terrain_flag,nx,ny,axisymm,nodex,nodey,myi,myj,viscosity, &
@@ -95,7 +95,7 @@
       real, intent(in), dimension(ibl:iel,jbl:jel) :: s10
       real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: rho,prs,ta
       real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: pi0,th0,th_in,ppa
-      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: uten,vten,wten,thten,qten,dten
+      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke,6) :: dpten
       real, intent(inout), dimension(ib:ie,jb:je,kb:ke,numq) :: qa
       real, intent(inout), dimension(ib:ie+1,jb:je,kb:ke) :: ua
       real, intent(inout), dimension(ib:ie,jb:je+1,kb:ke) :: va
@@ -354,7 +354,7 @@
 #ifdef _B4B01F
     !$acc update &
     !$acc host(pdata,xf,yf,zf,zh,sigma,sigmaf, &
-    !$acc      ua,va,wa,qa,ta,rho,prs,znt,qten,thten,zs)
+    !$acc      ua,va,wa,qa,ta,rho,prs,znt,dpten,zs)
 #else
     !JMD WARNING: Loop does not yet match CPU version.
     !$acc parallel default(present)
@@ -488,7 +488,7 @@
 
       !Project the feedback onto the grid, using ORIGINAL location
       call project_feedback(iflag,jflag,kflag,x3d0,y3d0,z3d0,sig3d0, &
-                            rhop0,taup0,rp0,tval,uten,vten,wten,qten,thten,dten, &
+                            rhop0,taup0,rp0,tval,dpten, &
                             xh,xf,yh,yf,zh,zf,sigma,sigmaf, &
                             sigdot,dvpdt1,dvpdt2,dvpdt3,drpdt,dtpdt,dmpdt, &
                             rhoval,part_grav1,part_grav2,part_grav3,pdata,np)
@@ -499,14 +499,13 @@
       stop 'FATAL error occurred in the index search BE_integration'
     endif
 
-    !Now that the tendency terms are computed, need to adjust the thten by the exner function:
 
     !$acc parallel default(present)
     !$acc loop gang vector collapse(3)
     do k=kb,ke
        do j=jb,je
           do i=ib,ie
-             thten(i,j,k) = thten(i,j,k)*pi0(i,j,k)
+             dpten(i,j,k,5) = dpten(i,j,k,5)*pi0(i,j,k)
           end do
        end do
     end do
@@ -1522,8 +1521,8 @@
       end subroutine interpolate_to_parcel
 
       subroutine project_feedback(iflag,jflag,kflag,x3d,y3d,z3d,sig3d, &
-                                  rhop0,taup0,rp0,tval,uten,vten,wten, &
-                                  qten,thten,dten,xh,xf,yh,yf,zh,zf,sigma,  &
+                                  rhop0,taup0,rp0,tval, &
+                                  dpten,xh,xf,yh,yf,zh,zf,sigma,  &
                                   sigmaf,sigdot,dvpdt1,dvpdt2,dvpdt3,  &
                                   drpdt,dtpdt,dmpdt,rhoval,part_grav1, &
                                   part_grav2,part_grav3,pdata,np)
@@ -1549,7 +1548,7 @@
       real, intent(in), dimension(ib:ie,jb:je,kb:ke) :: zh
       real, intent(in), dimension(kb:ke) :: sigma
       real, intent(in), dimension(kb:ke+1) :: sigmaf
-      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: uten,vten,wten,thten,qten,dten
+      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke,6) :: dpten
 
       integer, intent(in) :: np
       integer, intent(in) :: iflag,jflag,kflag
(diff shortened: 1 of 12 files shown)


src/solve3.F
```diff
@@ -951,67 +951,47 @@ subroutine solve3(nstep,rbufsz,num_soil_layers,                  &
...
-
           !$acc loop gang vector collapse(3)
           do k=kb,ke
           do j=jb,je
           do i=ib,ie
             dpten(i,j,k,1) = 0.0
             dpten(i,j,k,2) = 0.0
+            dpten(i,j,k,3) = 0.0
+            dpten(i,j,k,4) = 0.0
+            dpten(i,j,k,5) = 0.0
```
sjsprecious, line 962: these three variables refer to the previous `wten`, `uten` and `vten` variables. we may need to initialize the **ie+1**, **je+1** and **ke+1** index for corresponding variables, too or it does not matter?

src/solve3.F
```diff
@@ -262,7 +262,7 @@ subroutine solve3(nstep,rbufsz,num_soil_layers,                  &
       real, intent(inout), dimension(ibl:iel,jbl:jel) :: tmn,tml,t0ml,hml,h0ml,huml,hvml,tmoml
       real, intent(inout), dimension(ibp:iep,jbp:jep,kbp:kep,npt) :: pta,pt3d,ptten
       real, intent(inout), dimension(nparcelsLocal,npvals) :: pdata
-      real, intent(inout), dimension(ib:ie,jb:je,kb:ke,2) :: dpten
+      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke,6) :: dpten
```
sjsprecious, line 265: the dimension of `dpten` and `uten`, `vten`, `wten` is not exactly the same. For example, `uten` is **(ib:ie+1,jb:je,kb:ke)**. Does it matter to change the dimension bound?

src/solve3.F
```diff
@@ -1030,38 +1014,62 @@ subroutine solve3(nstep,rbufsz,num_soil_layers,                  &
...
+                                     ws31,ws32,wn31,wn32,reqs_w)
+            !----
 #endif
             !$omp parallel do default(shared) private(i,j,k)
             !$acc parallel loop gang vector collapse(3) default(present) private(i,j,k)
             do k=1,nk
-            do j=0,nj+1
-            do i=0,ni+1
+            do j=0,nj+2
+            do i=0,ni+2
```
sjsprecious, line 1066: As @johnmauff pointed out, the change of loop bound here looked suspicious.

src/droplet.F
```diff
@@ -94,13 +93,13 @@ subroutine droplet_driver(dt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,
       real, intent(in), dimension(kb:ke+1) :: sigmaf
       real, intent(in), dimension(ib:ie,jb:je) :: znt
       real, intent(in), dimension(ibl:iel,jbl:jel) :: s10
-      real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: rho,prs
+      real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: rho,prs,ta
       real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: pi0,th0,th_in,ppa
-      real, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: thten,qten
+      double precision, intent(inout), dimension(ib:ie,jb:je,kb:ke) :: uten,vten,wten,thten,qten,dten
```
sjsprecious: Again the potentially inconsistent dimension between `dten`, `uten`, `vten` and `wten`.

src/droplet.F
```diff
@@ -489,7 +488,7 @@ subroutine droplet_driver(dt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,mf,zs,
 
       !Project the feedback onto the grid, using ORIGINAL location
       call project_feedback(iflag,jflag,kflag,x3d0,y3d0,z3d0,sig3d0, &
-                            rhop0,taup0,rp0,tval,uten,vten,wten,qten,thten, &
+                            rhop0,taup0,rp0,tval,uten,vten,wten,qten,thten,dten, &
```
sjsprecious: @drichte2 So `uten`, `vten`, `wten`, `qten` and `thten` are still used in the **project_feedback** subroutine. Should them be replaced by `dten` as well?

src/solve3.F
drichte2: `dpten` is only allocated for `dpten(ib:ie,jb:je,kb:ke,6)` now. These indices encompass all of the grid points that could be affected by particles, so there is no need to extend to ie+1, je+1, or ke+1.

src/solve3.F
drichte2: I really need George's confirmation, since he apparently did things differently than what I was expecting, but the velocity feedbacks from droplets (`uten`, `vten`, `wten`) can completely fit within the ranges `ib:ie`, `jb:je`, `kb:ke`. Those extra ghost points are needed for other types of tendency terms, but not the droplet ones.

It does get a little confusing now: at the cm1.F and solve3.F levels, uten, vten, and wten have `ie+1`, `je+1`, and `ke+1` bounds, respectively. But inside droplet_driver, they can all be the same, corresponding to the size of dpten. We could perhaps consider changing the variable names `uten`, `vten`, and `wten` inside droplet.F?

src/solve3.F
drichte2: We need @george-bryan to confirm here. While it seems safe (i.e., it's within the bounds of these arrays), I'm not certain why it was changed.

src/droplet.F
drichte2: See above

src/droplet.F
drichte2: `dten` (not to be confused with `dpten`) is a new stats variable that was created to get a 3D grid of the droplet numbers. So this is just adding to it, not replacing any of the others.

sjsprecious: Thanks @drichte2 for your clarification. So should we use `dpten` instead of `uten`, `vten` and `wten` for the **project_feedback** subroutine?

drichte2: I was just about to revise my comment above about renaming the variables inside project_feedback. We could simply remove all references to `uten`, `vten`, `wten`, `qten`, `thten`, and `dten` and just use `dpten` as appropriate. Would that work?

sjsprecious: I think this will be great. I also suggest that we should replace all those `Xten` variables by `dpten` in CM1 to avoid confusion and potential bugs.

drichte2: I will make those changes, since I was already cleaning something up.

Your 2nd suggestion wouldn't work I think. In cm1.F and solve3.F, they all have different sizes. `dpten` would only be referring to the droplet feedback. That way, any reference to `Xten` inside cm1.F or solve3.F is coming from some other tendency, NOT the droplets.

sjsprecious: So you mean `uten`, `vten`, etc in cm1.F, solve3.F and droplet.F are actually different variables and not related to each other?

src/solve3.F
```diff
@@ -951,67 +951,47 @@ subroutine solve3(nstep,rbufsz,num_soil_layers,                  &
...
+          call comm_1s_tend_halo_GPU(dpten(ib,jb,kb,5))
 #else
           call bcs_tend_halo(dpten(ib,jb,kb,1))
           call bcs_tend_halo(dpten(ib,jb,kb,2))
 #endif
 #ifdef MPI
           if(timestats.ge.1) time_dpc=time_dpc+mytime()
 #endif
-          if( nout3d.ge.2 )then
+          if( nout3d.ge.6 )then
```
sjsprecious, line 993: where is the variable `nout3d` set? I only see it initialized to 0 in cm1.F

src/droplet.F
drichte2: In cm1.F, there are "ten" terms for all of the variables with some governing equation (`uten`, `vten`, ... `qvten`, `thten`, `ppten`, etc). These are supposed to be general and can account for any source/sink in any of the associated governing equations. Due to grid staggering, they don't all have the same size (hence the ie+1, je+1, etc), and so there is no convenient way to consolidate them all into a single "ten" array. 

So these are what are used consistently throughout cm1.F and solve3.F. The plan was to do the same thing with the droplets (hence why they were all passed in individually at first). But George configured the droplet feedbacks so that the "ten" arrays associated with droplets could all have the same size (again, I need to confirm with him how exactly he did this). This allowed him to consolidate them into a single `dpten` array. So physically, the various elements have the exact same meaning as `uten`, `vten`, etc. in cm1.F and solve3.F -- but specific to the droplets, these were computed in a way that allows them to be joined together into a single array. Hope that makes sense?

src/solve3.F
drichte2: Right -- it's set at the top of cm1.F. It's typically set to 0. The user can change this in order to output 3D variables if they want to analyze them. Otherwise the code ignores them.

src/droplet.F
```diff
@@ -354,7 +354,7 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
 #ifdef _B4B01F
     !$acc update &
     !$acc host(pdata,xf,yf,zf,zh,sigma,sigmaf, &
-    !$acc      ua,va,wa,qa,ta,rho,prs,znt,qten,thten,zs)
```
drichte2, line 358: @sjsprecious Please note that George appears to have forgotten to add `uten`, `vten`, `wten` here when he updated. This could be a reason why the GPU and CPU versions didn't match? I replaced these with `dpten`, which is hopefully correct.

johnmauff: This should not be a problem, as this is dead code that has not yet been eliminated.  Note that it is enclosed in the #ifdef _B4B01F parameter which is not set.

src/droplet.F
sjsprecious: Hi @drichte2 , thanks for your detailed explanation. That makes a lot of sense to me now.

One more question: If we use `dpten` in the project_feedback subroutine and probably throughout the droplet subroutine, do we need to map `dpten` back to `uten`, `vten`, `wten`, `thten` and `qten` in the cm1.F and solve3.F so that they can be populated in the other calculations?

src/solve3.F
sjsprecious: Oh, I see. This is just for output purpose, not affecting the results?

src/droplet.F
drichte2: Gotcha, didn't notice that

src/droplet.F
drichte2: No, these don't need to be mapped back. They get used immediately after droplet_driver inside solve3.F, inside the `IF ( two_way_coupling )` block. Here the droplet tendencies are "felt" by the velocity, temperature, and humidity directly by `dpten`, and afterwards `dpten` is no longer needed (except for stats).

src/solve3.F
drichte2: Correct

src/droplet.F
sjsprecious: Thanks for the clarification!

src/droplet.F
drichte2: @johnmauff you both obviously have more expertise than I do, but looking back through, I never see `dpten` being updated on either host or device. How can this be the case? Am I missing it somewhere, or perhaps this is somehow automatically accounted for with the `acc atomic update` above each of the `dpten` updates in `project_feedback`? If not, I don't see how these GPU atomic calculations for `dpten` ever make their way back to the CPU before being used inside solve3.F

johnmauff (approved):

PR #124, opened 2023-07-03
Eliminated duplicate comm subroutines

johnmauff:
This is the second large PR that involves the deletion of duplication communication subroutines.  Additionally, the _GPU tag has also been deleted.  Several issues with dead code were resolved in the turb.F file.  The correctness statistics are slightly better than  PR #121 but slightly worse than PR #123

ifort CPU versus nvhpc GPU
=====================

Metrics
--------
39 stat variables are identical.
42 stat variables show a mean relative difference > 1e-06
5 stat variables show a mean relative difference <= 1e-06

0th-order
----------
3 stat variables are identical.
7 stat variables show a mean relative difference > 1e-06
2 stat variables show a mean relative difference <= 1e-06

1st-order
----------
254 stat variables are identical.
2 stat variables show a mean relative difference > 1e-06
0 stat variables show a mean relative difference <= 1e-06

PR #125, opened 2023-07-11
Added ufmean to the droplet_diag

drichte2:
Adds a few more items to droplet_diag

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -502,7 +502,7 @@
     do k=kb,ke
        do j=jb,je
           do i=ib,ie
-             dpten(i,j,k,5) = dpten(i,j,k,5)*pi0(i,j,k)
+             dpten(i,j,k,1) = dpten(i,j,k,1)*pi0(i,j,k)
           end do
        end do
     end do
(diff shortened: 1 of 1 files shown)


sjsprecious:
The verification result on Gust with `namelist_ASD.verify` namelist is similar with that from PR #124.

# 2 MPI ranks (CPU, intel) versus 2 A100 GPU (nvhpc)

## Metrics

38 stat variables are identical

## 0th-order

4 stat variables are identical

## 1st-order

254 stat variables are identical

src/droplet.F
```diff
@@ -502,7 +502,7 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
     do k=kb,ke
        do j=jb,je
           do i=ib,ie
-             dpten(i,j,k,5) = dpten(i,j,k,5)*pi0(i,j,k)
+             dpten(i,j,k,1) = dpten(i,j,k,1)*pi0(i,j,k)
```
sjsprecious, line 505: Is this a bug fix or just due to a reorder of `dpten` index?

sjsprecious (commented):
Thanks @drichte2 for issuing this PR. It looks good to me and the verification result looks reasonable, too. I just have one clarification question.

src/droplet.F
drichte2: Bug fix! George pointed it out last week

sjsprecious: Thanks for the clarification.

sjsprecious (approved):

johnmauff (approved):
Looks good to me.

Issue #120, opened 2023-06-21
Improved restart functionality for CM1


sjsprecious:
The current restart functionality does not work for large number of droplets (>200 M) due to the restriction of `int` amount of data that can be exchanged through `mpi_bcast`.

Based on the meeting with George/David/John, we will assume that the restart run uses the same number of MPI ranks as the initial run so that we could send a known size of buffer to each MPI rank specifically, without using the `mpi_bcast` any more.

sjsprecious:
Closed via PR https://github.com/george-bryan/CM1/pull/127.

PR #127, opened 2023-07-17
Fix the broken restart functionality

sjsprecious:
This PR fixes the broken restart functionality reported by @drichte2 , where he tried to restart a run with > 100 M droplets.

The underlying reason is that the original implementation uses the **MPI_BCAST** interface and it could only send/receive `int` number of data, which will be exceeded if there are > 100 M droplets.

The new implementation uses **MPI_SEND** and **MPI_RECV** interfaces instead. We assume that the same PE layout is used for the initial and restart runs. We write the number of active droplets from each MPI rank to the restart file and the root MPI rank knows how many active droplets will be sent to each MPI rank during the restart run later.

I verified that the initial run (use `namelist_ASD.restC`) and restart run (use `namelist_ASD.restA` and `namelist_ASD.restB`) match for the total droplet number (> 100 M) when I used 4 MPI ranks + 4 GPUs on Gust.

Fix https://github.com/george-bryan/CM1/issues/120.

### DIFF
--- a/src/restart.F
+++ b/src/restart.F
@@ -3728,11 +3728,6 @@
             if ( iprcl.eq.1 ) then
 #ifdef MPI
                call mpi_bcast(nparcels_per_mpi,numprocs,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
-               if ( myid .ne. 0 ) then
-                  if ( .not. allocated(tmp_buf) ) then
-                     allocate( tmp_buf(nparcels_per_mpi(myid+1),npvars) )
-                  end if
-               end if
                if ( myid .eq. 0 ) then
                   do i = 1, numprocs - 1
                      allocate( tmp_buf(nparcels_per_mpi(i+1),npvars) )
@@ -3746,6 +3741,7 @@
                   end do
                else
                   tag = 98765 + myid
+                  allocate( tmp_buf(nparcels_per_mpi(myid+1),npvars) )
                   call mpi_recv(tmp_buf,nparcels_per_mpi(myid+1)*npvars, &
                                 MPI_REAL,0,tag,MPI_COMM_WORLD, &
                                 MPI_STATUS_IGNORE,ierr)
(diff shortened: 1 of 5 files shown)


src/restart.F
```diff
@@ -3697,59 +3708,74 @@ subroutine read_restart(nstep,srec,sirec,urec,vrec,wrec,nrec,mrec,prec,      &
...
-          IF( iprcl.eq.1 )THEN
+               end if
+            end if
+            if ( iprcl.eq.1 ) then
 #ifdef MPI
-            call MPI_BCAST(nparcelsRest,1,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
-            if ( .not. allocated(rbuf) ) allocate( rbuf(nparcelsRest,npvars) )
-            call MPI_BCAST(rbuf,npvars*nparcelsRest,MPI_REAL,0,MPI_COMM_WORLD,ierr)
+               call mpi_bcast(nparcels_per_mpi,numprocs,MPI_INTEGER,0,MPI_COMM_WORLD,ierr)
+               if ( myid .ne. 0 ) then
```
johnmauff: @sjsprecious Why do you separate the allocation of tmp_buf from its use in an MPI_recv call?  It seems like it would make the code easier to read if they were in the same if block

sjsprecious: Thanks for catching it! Just fix it based on the suggestion.

johnmauff (approved):
Looks good to me once you merge in changes from the gpu-opt branch,

drichte2:
I also checked this using a version of what I've been running. Restarted just fine with > 100M particles.

drichte2 (approved):

sjsprecious:
> Looks good to me once you merge in changes from the gpu-opt branch,

Thanks @johnmauff . Just merge the latest `gpu-opt` branch.

sjsprecious:
> I also checked this using a version of what I've been running. Restarted just fine with > 100M particles.

Thanks @drichte2 .

PR #128, opened 2023-07-20
Count prefix

johnmauff:
This PR eliminates the last remaining significant data movement associated with the droplet code.  This is achieved using the count_prefix() operator which counts the number of occurrences of a particular value in a GPU-resident array using a parallel prefix algorithm.  Note that a CUDA function in the CUDA tensorex module provides this capability.  An example use of this operator is as follows

     use tensorex
     locind = [1 2 3 4 undefined 5 6 7 undefined 8]
     C = COUNT_PREFIX(mask = (locind .eq. undefined), EXCLUSIVE=.true.)
     ! would generate the array 
     ! C = [0 0 0 0 0 1 1 1 1 2]

The build of CM1 needs to be slightly modified.  On Derecho, this is achieved by loading of the module 'cutensor'.  On Gust the build is not quite as clean and requires the setting of two environment variables.

export LIBRARY_PATH=$LIBRARY_PATH:/glade/u/apps/common/22.12/spack/opt/spack/nvhpc/23.3/Linux_x86_64/23.3/math_libs/12.0/targets/x86_64-linux/lib
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/glade/u/apps/common/22.12/spack/opt/spack/nvhpc/23.3/Linux_x86_64/23.3/math_libs/12.0/targets/x86_64-linux/lib

The verification results are similar to the previous PR, and are as follows:

Metrics:
======

Oth-order diagnostics:
=================

1st-order diagnostics 
================

### DIFF
--- a/src/comm_droplet.F
+++ b/src/comm_droplet.F
@@ -680,12 +680,14 @@ contains
       integer :: n_idx,s_idx,e_idx,w_idx,nw_idx,ne_idx,sw_idx,se_idx
       integer np
       logical, dimension(nparcelsLocalActive) :: mskN,mskS,mskW,mskE,mskNW,mskNE,mskSW,mskSE
-      integer, dimension(nparcelsLocalActive) :: CidxN,CidxS,CidxW,CidxE,CidxNW,CidxNE,CidxSW,CidxSE
-      integer, allocatable :: Depart_ind2(:)
+      integer :: Cidx(nparcelsLocalActive)
       integer :: numDepart,numerrors
+      logical, parameter :: verbose = .false.
+      !!$acc declare present(ptrDepart,Depart,Depart_ind,pdata_neighbor)
+      !$acc declare present(Depart_ind,pdata_neighbor)
 
       numDepart=SUM(Depart)
-      if(numDepart.gt.0) allocate(Depart_ind2(numDepart))
+!      if(numDepart.gt.0) allocate(Depart_ind2(numDepart))
 
       n_idx  = ptrDepart(inorth)
       s_idx  = ptrDepart(isouth)
@@ -695,15 +697,14 @@ contains
       ne_idx = ptrDepart(ine)
       sw_idx = ptrDepart(isw)
       se_idx = ptrDepart(ise)
-      if(numDepart.gt.0) then
+      if(verbose .and. (numDepart.gt.0)) then
          write(*,110) myid,'Depart:',Depart
          write(*,110) myid,'ptrDepart:',ptrDepart
       endif
 
-!$acc data copyin(ptrDepart,Depart,Depart_ind2,pdata_neighbor) &
-!$acc create(mskN,mskS,mskW,mskE,mskNW,mskNE,mskSE,mskSW, &
-!$acc        CidxN,CidxS,CidxW,CidxE,CidxNW,CidxNE,CidxSW,CidxSE)
+#ifdef _COUNTPREFIX
 
+!$acc data create(mskN,mskS,mskW,mskE,mskNW,mskNE,mskSE,mskSW,Cidx)
 
       !$acc parallel loop gang vector default(present)
       do np=1,nparcelsLocalActive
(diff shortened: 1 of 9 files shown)


src/comm_droplet.F
```diff
@@ -660,23 +660,190 @@ end subroutine setupIndexPointers
...
+         !call GatherIndices(inorth,ptrDepart,Cidx,pdata_neighbor,Depart_ind)
+         Cidx  = count_prefix(mask=(pdata_neighbor .eq. inorth),exclusive=.true.)
+         !$acc parallel loop gang vector default(present)
+         do np=1,nparcelsLocalActive-1
+            if(Cidx(np) .ne. Cidx(np+1)) then
+               Depart_ind(n_idx+Cidx(np))=np
+            endif
+         enddo
+         ! special treament for the last value in the array
+         !$acc parallel default(present) num_gangs(1) num_workers(1)
```
sjsprecious: I think using `!$acc serial default(present)` is probably more straightforward?

src/comm_droplet.F
```diff
@@ -686,35 +853,84 @@ subroutine setupDepartDroplet(Depart_ind,ptrDepart,pdata_neighbor)
       sw_idx = ptrDepart(isw)
       se_idx = ptrDepart(ise)
       do np=1, nparcelsLocalActive
-        if(pdata_neighbor(np) .eq. mynorth) then
+        if(pdata_neighbor(np) .eq. inorth) then
```
sjsprecious, line 747: what is the difference between `mynorth` and `inorth`?

src/comm_droplet.F
```diff
@@ -660,23 +660,190 @@ end subroutine setupIndexPointers
...
+            if(Cidx(np) .ne. Cidx(np+1)) then
+               Depart_ind(s_idx+Cidx(np))=np
+            endif
+         enddo
+         ! special treament for the last value in the array
+         !$acc parallel default(present) num_gangs(1) num_workers(1)
+         if(pdata_neighbor(nparcelsLocalActive).eq.isouth) then
+            Depart_ind(s_idx+Cidx(nparcelsLocalActive)) = nparcelsLocalActive
+         endif
+         !$acc end parallel
```
sjsprecious: This code structure has been repeated eight times here. Can we write a subroutine interface instead to reduce the potential code bug due to copy & paste?

src/droplet.F
```diff
@@ -449,22 +457,29 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
                           neighbor,num_fallout,errcode)
 
 #ifdef MPI
-      pdata_neighbor(np) = neighbor 
       if(neighbor .eq. mynorth) then 
+         pdata_neighbor(np) = inorth
```
sjsprecious, line 461: why we need this change instead of just using `pdata_neighbor(np) = neighbor`?

src/droplet.F
```diff
@@ -200,8 +210,7 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
 !     (for the parcel subroutines only!)
 
 #ifdef MPI
-      !$acc data copyout (pdata_neighbor) &
-      !$acc      copyin  (randomNumbers,randomNumbers%state)
+      !$acc data create(pdata_neighbor,msk)
```
sjsprecious, line 213: why we do not need `!$acc copyin (randomNumbers,randomNumbers%state)`?

src/droplet.F
```diff
@@ -528,8 +533,9 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
     Depart(isw)    = numDepartSW
     Depart(ise)    = numDepartSE
 
-    numDepart=SUM(Depart)
-    if(numDepart .gt. 0) allocate(Depart_ind(numDepart))
+     numDepart=SUM(Depart)
+     if(numDepart .gt. 0) allocate(Depart_ind(max(numDepart,padding)))
```
sjsprecious: why not just `allocate(Depart_ind(numDepart))`?

src/droplet.F
```diff
@@ -545,20 +551,62 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
      allocate(holes_ind(numHoles))
+!$acc enter data create(holes_ind)
 
-     i=0
 
+     !$acc parallel loop gang vector default(present) 
+     do np=1,nparcelsLocal
+       msk(np) = (pdata_locind(np,1) .eq. undefined_index) .and. &
+                 (pdata_locind(np,2) .eq. undefined_index)
+     enddo
```
sjsprecious, line 560: For consistency, add `!$acc end parallel`.

src/droplet.F
```diff
@@ -545,20 +551,62 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
+
+#if (defined(_COUNTPREFIX) && defined(_OPENACC))
+
+     !$acc data create(Cindx)
+     !$acc parallel loop gang vector default(present)
      do np = 1, numHoles
        holes_ind(np) = undefined_index
      end do
```
sjsprecious, line 569: For consistency, add `!$acc end parallel`.

src/droplet.F
```diff
@@ -545,20 +551,62 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
+     ! C = [0 0 0 0 0 1 1 1 1 2]
+     Cindx = count_prefix(mask=msk,exclusive=.true.)
+
+     ! collect the location of the special values into a smaller array
+     !$acc parallel loop gang vector default(present)
+     do np=1,nparcelsLocal-1
+        if(Cindx(np).ne.Cindx(np+1)) then 
+           holes_ind(Cindx(np)+1)=np
+        endif
+     enddo
```
sjsprecious: For consistency, add `!$acc end parallel`.

src/droplet.F
```diff
@@ -593,19 +641,21 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
        allocate(backfill_ind(numBackfill))
+       !$acc enter data create(backfill_ind)
        if ((nparcelsLocalActive - numBackfill) .eq. 0) then
           ! Set backfill_ind to 1 to handle an exception case, 
           ! where all the active droplets fall out or leave
           ! the current MPI domain
+          !$acc parallel loop gang vector default(present)
           do i = 1, numBackfill
              backfill_ind(i) = 1
           end do
```
sjsprecious, line 659: For consistency, add `!$acc end parallel`.

src/droplet.F
```diff
@@ -593,19 +641,21 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
        else
+          !$acc parallel num_gangs(1) num_workers(1) default(present)
```
sjsprecious, line 661: Probably using `!$acc serial default(present)` is more straightforward?

src/droplet.F
```diff
@@ -694,13 +746,18 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
 
     ! free up the memory for temporary variables
 
-    deallocate(holes_ind)
-    if(allocated(Depart_ind)) deallocate(Depart_ind)
-    if(allocated(backfill_ind)) deallocate(backfill_ind)
+     !$acc exit data delete(holes_ind)
+     deallocate(holes_ind)
+
+     !$acc exit data delete(Depart_ind)
```
sjsprecious: should this line be moved inside the `if` block below?

src/droplet.F
```diff
@@ -694,13 +746,18 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
...
+     if(allocated(Depart_ind)) deallocate(Depart_ind)
+
+     !$acc exit data delete(backfill_ind)
```
sjsprecious: should this line be moved inside the if block below?

src/comm_droplet.F
```diff
@@ -649,9 +649,9 @@ subroutine setupIndexPointers(Cnt,ptrCnt)
 
       ptrCnt(inorth) = 1
       ptrCnt(isouth) = ptrCnt(inorth) + Cnt(inorth)
-      ptrCnt(ieast)  = ptrCnt(isouth) + Cnt(isouth)
-      ptrCnt(iwest)  = ptrCnt(ieast)  + Cnt(ieast)
-      ptrCnt(inw)    = ptrCnt(iwest)  + Cnt(iwest)
+      ptrCnt(iwest)  = ptrCnt(isouth) + Cnt(isouth)
+      ptrCnt(ieast)  = ptrCnt(iwest)  + Cnt(iwest)
+      ptrCnt(inw)    = ptrCnt(ieast)  + Cnt(ieast)
```
sjsprecious, line 654: is there a particular reason to switch the order here?

src/comm_droplet.F
```diff
@@ -686,35 +853,84 @@ subroutine setupDepartDroplet(Depart_ind,ptrDepart,pdata_neighbor)
...
+ 110     format('myid: ',i4,' ',a10,' ',8(i8,1x))
 
      end subroutine setupDepartDroplet
 
+#if (defined(_COUNTPREFIX) && defined(_OPENACC))
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+
+     subroutine GatherIndices(dir,ptrDepart,Cidx,pdata_neighbor,Depart_ind)
```
sjsprecious: Where is this subroutine being called?

src/comm_droplet.F
johnmauff: @sjsprecious  I actually did write a subroutine interface to execute these calculations, it is called GatherIndices.  Unfortunately, it suffers from a runtime error. I would appreciate it if you could take a look at it and see if you can see why it is not working.

src/comm_droplet.F
johnmauff: The variable "mynorth" is the MPI rank of the northern neighbor, while 'inorth' is the index of the northern neighbor.

src/droplet.F
johnmauff: @sjsprecious Good question.  I needed to redefine the meaning of pdata_neighbor because it allows us to identify particles that are on the border to be treated properly and to perform this logic in one place instead of multiple places.  It also eliminates the need to create individual mask logic variables so that we can call count_prefix.

src/droplet.F
johnmauff: The randomNumbers is no longer used in the droplet_driver subroutine.  It has been moved to the droplet injection code.

src/comm_droplet.F
johnmauff: The order of the droplets in the Depart_ind was not consistent with the order of the integer direction pointers (inorth,isouth, iwest,ieast,...)  Making it consistent fixed a very vexing correctness bug.

src/comm_droplet.F
johnmauff: If you want to figure out why this subroutine is failing at runtime, then we could replace a large amount of the duplicate code in setupDepartDroplet.

src/droplet.F
johnmauff: Good catch.  I will add it to the if block.

src/droplet.F
johnmauff: Good catch.  I will add it to the if block.

src/comm_droplet.F
sjsprecious: Ah, I see. That is why the subroutine is defined but never called in this PR. I will take a look at it.

src/droplet.F
sjsprecious: Thanks for the clarification.

src/droplet.F
sjsprecious: Gotcha!

src/comm_droplet.F
sjsprecious: Thanks.

src/comm_droplet.F
johnmauff: Good point, I have updated the source code

src/droplet.F
johnmauff: Just changed the source code.

src/droplet.F
johnmauff: This is related to the lazy garbage collection and memory leakage in OpenACC.  If the size of the array in this case Depart_ind increases in successive calls then it will allocate a new array that is large enough.  Unfortunately, it does not immediately deallocate the old small array on the GPU device.  So this results in a memory leak until a garbage collection code deallocates all the unused memory.  I therefore choose to allocate a much larger array initially in an attempt to eliminate memory leaks.

src/droplet.F
johnmauff: @sjsprecious Actually there is no consistency throughout CM1 regarding the use of '$acc end parallel'.  I prefer to not include it so as to minimize the actual number of $acc directives present in the code.  I don't think it adds anything with respect to clarity.

sjsprecious: I see. But I still think `!$acc end parallel` is helpful for clarity, especially when we have multiple kernels consecutively or have a very long kernel.

src/droplet.F
sjsprecious: Oh, I see. Thanks for the clarification. It sounds like `!$acc exit data delete (Depart_ind)` does not work as expected. Probably we can discuss with the NVIDIA team next time.

src/comm_droplet.F
sjsprecious: I am still unclear about why this change is necessary. Can you provide an example?

src/droplet.F
sjsprecious: @johnmauff , if I used the `namelist_ASD.verify` namelist, what value of`timax` should I use to trigger the memory leak error without the code modification here?

src/comm_droplet.F
johnmauff: Probably best to discuss over a video call.  I made this change based on my understanding of something you told me a number of months ago about a very subtle bug.

src/comm_droplet.F
sjsprecious: @johnmauff I have revised the code to make the `GatherIndices` subroutine work, at the cost of creating `Cidx` on the GPU every time this subroutine is called. I am not sure how large the impact is on the GPU performance.

However, this revision does not seem necessary to me. Since `count_prefix` is introduced in `nvhpc/23.5` (https://docs.nvidia.com/hpc-sdk/compilers/cuda-fortran-prog-guide/#cfref-fort-mods-cutensorex-cpfx), I suspect that this is just another compiler bug. Feel free to revert my commit if you do not like this change.

src/droplet.F
sjsprecious: This no longer seems an issue and we have removed the `max` function.

sjsprecious (approved):

PR #129, opened 2023-08-04
Enable long integer type for large droplet count

### LINKED ISSUE #126: Use long integer type for the droplet count
According to David, it will be good to set all the variables related to the droplet count to long integer type so that CM1 can simulate more than 2 billions droplets.

sjsprecious:
This PR allows CM1 to simulate number of droplets that can not be represented by the default integer type. The code now worked for 3 billions droplets.

I declared many droplet count variables as `integer(i8)` type when it was necessary. As a result, the `nparcelsInit` and `nparcelsMax` namelist variables need to be expanded explicitly.

The current I/O implementation did not work for 3 billions droplets as it would exceed the memory limitation on Derecho. As a temporary solution, we only wrote out a fraction of total droplets to the binary file.

The verification results are similar to the previous PR (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs):

# Metrics
41 stat variables are identical.

# 0th-order diagnostics
7 stat variables show a mean relative difference > 1e-06

# 1st-order diagnostics

---

Fix #126

### DIFF
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -2276,7 +2276,7 @@ end module cuda_rt
 #endif
             ! number of droplets read from / write to a restart file
             ! initialize its value as "nparcelsInit"
-            nparcelsRest = int(nparcelsInit, i8)
+            nparcelsRest = nparcelsInit
             if(dowr) write(outfile,*) 'PDATA: padding,nparcelsLocal,nparcelsMax',padding,nparcelsLocal,nparcelsMax
          end if
       else   ! if no parcels
--- a/src/cm1.F
+++ b/src/cm1.F
@@ -2276,7 +2276,7 @@ end module cuda_rt
 #endif
             ! number of droplets read from / write to a restart file
             ! initialize its value as "nparcelsInit"
-            nparcelsRest = nparcelsInit
+            nparcelsRest = int(nparcelsInit, i8)
             if(dowr) write(outfile,*) 'PDATA: padding,nparcelsLocal,nparcelsMax',padding,nparcelsLocal,nparcelsMax
          end if
       else   ! if no parcels
--- a/src/comm_droplet.F
+++ b/src/comm_droplet.F
@@ -905,7 +905,7 @@ contains
 !ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 !ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 
-     subroutine CollectDropInfo ( pdata, rbuf, nparcels_per_mpi, nparcelstot )
+     subroutine CollectDropInfo ( pdata, rbuf, nparcels_per_mpi, nparcelsout )
 
      ! This subroutine will collect the "pdata" information from each
      ! MPI rank and assemble them into a single big array "rbuf" for 
@@ -913,7 +913,6 @@ contains
      ! It works for > 2 billions droplets that can not be represented 
      ! by a normal integer type of length
 
-     use constants, only : i8
      use input, only : myid,numprocs,ierr,nparcelsLocalActive, &
                        nparcelsLocal,npvals
      use mpi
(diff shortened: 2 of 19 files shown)


src/droplet.F
```diff
@@ -862,7 +862,8 @@ subroutine droplet_driver(dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh,rvh,yf,zh,mh,rmh,zf,
       end if    ! if floor(mtime/drop_inject_elapse) .ne. floor((mtime+dt)/drop_inject_elapse) 
       !This is helpful to see in the output, but not absolutely necessary:
 #ifdef MPI
-      call mpi_allreduce(nparcelsLocalActive,np_tmp,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
+      call mpi_allreduce(real(nparcelsLocalActive),np_tmp,1,mpi_real, &
```
johnmauff: This seems like a problematic solution.  So MPI_INTEGER8 is not supported in the current MPI implementation?  Is there a way to use the C version of 64-bit integers?

src/cm1.F
```diff
@@ -2276,7 +2276,7 @@ program cm1
 #endif
             ! number of droplets read from / write to a restart file
             ! initialize its value as "nparcelsInit"
-            nparcelsRest = nparcelsInit
+            nparcelsRest = int(nparcelsInit, i8)
```
johnmauff: What are the sizes of nparcelsRest and nparcelsInit.  It seems like they should both be integer(i8).  If so what does there need to be an explicit cast here?

src/comm_droplet.F
```diff
@@ -858,7 +857,126 @@ subroutine update_new_droplet ( dir_idx, Arrive, ptrArrive, holes_ind, pdata, dr
...
      end do
 
-  end subroutine update_new_droplet
+     end subroutine update_new_droplet
+
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+!ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+
+     subroutine CollectDropCount ( nparcels_per_mpi, nparcelstot )
```
johnmauff, line 866: I don't understand why this is present in this PR.  I thought this PR was focused on supporting supporting greater than 2B droplets.  If this is a general I/O improvement than that should be a different PR.

src/droplet.F
```diff
@@ -3265,10 +3268,12 @@ subroutine droplet_injection(pdata,xf,yf,dt,time_since_inject,mtime,my_reintro)
 
 #ifdef MPI
       !For deciding whether to inject more:
-      call mpi_allreduce(my_reintro+nparcelsLocalActive,tot_after_reintro,1,mpi_int,mpi_sum,mpi_comm_world,ierr)
+      call mpi_allreduce(real(my_reintro+nparcelsLocalActive), &
```
johnmauff: Same comment as above.  Is there no way to get 8-byte integers working?

src/input.F
```diff
@@ -158,8 +159,10 @@ MODULE input
               bl_mynn_mixlength,bl_mynn_edmf,                                 &
               bl_mynn_edmf_mom,bl_mynn_edmf_tke,bl_mynn_mixscalars,           &
               bl_mynn_cloudmix,bl_mynn_mixqt,initflag,spp_pbl,bl_mynn_output, &
-              nutk,nvtk,nwtk,nparcelsLocal,nparcelsLocalActive,nparcelsMax,   &
-              nparcelsRest
+              nutk,nvtk,nwtk,nparcelsLocal,nparcelsLocalActive
+
+      real :: nparcelsInit,nparcelsMax
```
johnmauff: We need to find a better solution here than to turn nparcelsInt and nparcelsMax into real.

johnmauff (changes requested):
@sjsprecious.  This PR has several things that I am uncomfortable with. The first one is replacing integers with reals.  I think we need to find a different solution here.  The other one is why are you mixing what appears to be an I/O improvement with this PR for long integers.

src/droplet.F
sjsprecious: Revised as suggested.

src/cm1.F
sjsprecious: Revised as suggested and all droplet counts are `integer(i8)` type now.

src/comm_droplet.F
sjsprecious: This interface is temporarily used for now. May not be needed for PIO in the future.

src/droplet.F
sjsprecious: Revised as suggested.

src/input.F
sjsprecious: Revised as suggested.

johnmauff (approved):

drichte2:
Can the instances of nparcelsLocal and nparcelsLocalActive be corrected in writeout_nc.F?

sjsprecious:
> Can the instances of nparcelsLocal and nparcelsLocalActive be corrected in writeout_nc.F?

Thanks @drichte2 . I just made some changes to fix the variable type issue for the NETCDF output.

drichte2 (approved):
Thanks for changing the things with tnumpart. Looks good to me

PR #130, opened 2023-08-04
Inject every time step

drichte2:
Modified a few things to make it straightforward to inject droplets every time step. Do this by setting "droplet_inject_elapse" to a negative number in the namelist file. This is good for the ASD runs, but for other applications it might be necessary to wait longer than a single time step to inject.

### DIFF
--- a/src/solve1.F
+++ b/src/solve1.F
@@ -416,7 +416,6 @@
       else
         tqnudge = .false.
       endif
-      tqnudge = .false.
 
 
       timenudge = mtime+dt
--- a/src/solve1.F
+++ b/src/solve1.F
@@ -416,6 +416,7 @@
       else
         tqnudge = .false.
       endif
+      tqnudge = .false.
 
 
       timenudge = mtime+dt
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -143,6 +143,8 @@
 
       integer :: values(8),local_seed
 
+      integer :: doinject  !Do a call to droplet_inject at this time step?
+
       type(randomNumberSequence) :: randomNumbers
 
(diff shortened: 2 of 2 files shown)


src/solve1.F
```diff
@@ -416,6 +416,7 @@ subroutine solve1(nstep,rbufsz,num_soil_layers,                  &
       else
         tqnudge = .false.
       endif
+      tqnudge = .false.
```
johnmauff: @drichte2 This seems problematic to turn tqnudge off by default.  Is this valid for all uses of CM1 or just for CM1 running testcase == 10 and with droplets?

sjsprecious: So this setting will always set `tqnudge` to `False`. Is this intended? If so, we probably should remove the `if` block above.

johnmauff (commented):
We need to have to correctness verification results.

sjsprecious (commented):
Thanks @drichte2 for submitting this PR. I have a minor clarification question before I run the verification test.

drichte2:
This was a manual thing I did to configure the ASD runs like I wanted them. `testcase==10` turns on nudging by default, which is what I am overriding here. I suppose one way is to define a new `testcase` number which is identical except for this point, but this seemed easier for these specific runs.

I'll take it out of the PR and only use it in my local version

sjsprecious:
The verification results are similar to the previous PR (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs):

# Metrics

# 0th-order diagnostics

# 1st-order diagnostics

sjsprecious (approved):
Looks good to me.

johnmauff:
Thanks for the changes.  They look good to me.

PR #131, opened 2023-08-21
Improve the droplet ownership check

sjsprecious:
This PR:

- improved the droplet initialization in `init3d.F` suggested by @drichte2 . Instead of looping over the `nparcelsInit` droplets for each MPI rank, now each MPI rank only loops over `nparcelsInit/numprocs` droplets and they are all active. Thus there is no need to check the droplet ownership and we just need to change the x/y range to the boundary of each MPI rank before initialization. In this way, no more droplets will fall on the boundary between tiles.

- improved the droplet ownership check in the `droplet.F` code. We first moved a huge code block with several logical checks into a separate subroutine. Then we updated the ownership check logics to treat all the droplets on the boundary of a MPI rank as `local` (i.e., no MPI communication is needed).

The verification results are slightly worse than the previous PR #129 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs):

# Metrics

# 0th-order diagnostics

# 1st-order diagnostics

### DIFF
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -3888,8 +3888,7 @@
       !$acc routine seq
 
       use input, only : ni,nj,ib,ie,jb,je,ngxy,nx,ny,dx,dy,mywest, &
-                        mysw,mynw,myeast,myse,myne,mysouth,mynorth, &
-                        maxx,maxy
+                        mysw,mynw,myeast,myse,myne,mysouth,mynorth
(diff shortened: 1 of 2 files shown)


drichte2 (approved):
Looks good to me

johnmauff (approved):
This PR looks good to me as well

PR #133, opened 2023-09-19
Modifying th10 and q10

drichte2:
Added code to compute th10 and q10 inside of sfcphys.F. Note that this calculation is very crude, and will shortly be replaced with direct interpolation. The passing of th10 and q10 needed to be done, as did the creation of a new testcase=15 option.

### DIFF
--- a/src/comm_droplet.F
+++ b/src/comm_droplet.F
@@ -813,12 +813,11 @@ contains
         !$acc end parallel
 
         ! special treament for the last value in the array
-        !$acc update self(pdata_neighbor(nparcelsLocalActive), &
-        !$acc             Cidx(nparcelsLocalActive))
+        !$acc kernels default(present) 
         if(pdata_neighbor(nparcelsLocalActive).eq.dir) then
            Depart_ind(ptrdir+Cidx(nparcelsLocalActive)) = nparcelsLocalActive
         endif
-        !$acc update device(Depart_ind(ptrdir+Cidx(nparcelsLocalActive)))
+        !$acc end kernels
 
         !$acc end data
 
--- a/src/comm_droplet.F
+++ b/src/comm_droplet.F
@@ -813,11 +813,12 @@ contains
         !$acc end parallel
 
         ! special treament for the last value in the array
-        !$acc serial default(present)
+        !$acc update self(pdata_neighbor(nparcelsLocalActive), &
+        !$acc             Cidx(nparcelsLocalActive))
         if(pdata_neighbor(nparcelsLocalActive).eq.dir) then
            Depart_ind(ptrdir+Cidx(nparcelsLocalActive)) = nparcelsLocalActive
         endif
-        !$acc end serial 
+        !$acc update device(Depart_ind(ptrdir+Cidx(nparcelsLocalActive)))
 
         !$acc end data
 
--- a/README.namelist
+++ b/README.namelist
@@ -191,8 +191,10 @@
            settings. 
 
     11 =   SAS convective PBL
+    12 =   IB stuff?
+    13 =   ??
 
-    12 =   Hurricane boundary layer with LES laden with spray
+    15 =   Hurricane boundary layer with LES laden with spray
            - Based on Richter & Wainwright (2023, GRL)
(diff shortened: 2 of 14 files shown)


sjsprecious:
The verification results are much worse than the previous PR https://github.com/george-bryan/CM1/pull/131 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with testcase=15:

# Metrics
60 stat variables show a mean relative difference > 1e-06

# 0th-order diagnostics
2 stat variables are identical.
10 stat variables show a mean relative difference > 1e-06

# 1st-order diagnostics
128 stat variables are identical.
128 stat variables show a mean relative difference > 1e-06

---

There should be a bug in the GPU code related to the new changes.

sjsprecious:
After updating David's and my changes, the verification results are much improved and similar to the previous PR https://github.com/george-bryan/CM1/pull/131, except the stat metrics output (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with testcase=15:

# Metrics

30 stat variables are identical.
53 stat variables show a mean relative difference > 1e-06

# 0th-order diagnostics

# 1st-order diagnostics

src/comm_droplet.F
```diff
@@ -813,11 +813,12 @@ subroutine GatherIndices(dir,ptrDepart,pdata_neighbor,Depart_ind)
         !$acc end parallel
 
         ! special treament for the last value in the array
-        !$acc serial default(present)
+        !$acc update self(pdata_neighbor(nparcelsLocalActive), &
+        !$acc             Cidx(nparcelsLocalActive))
```
johnmauff: I don't understand why you are updating the CPU here.  The pdata_neighbour and Cidx array are never used by the CPU.

sjsprecious: So this is the key part where the error comes from. After doing a few tests and reading the OpenACC manual carefully, I realize that `!$acc parallel` and `!$acc serial` must be followed by a `do` loop immediately. If we are just updating one element of an array or a scalar variable (i.e., no `do` loop here), using `!$acc parallel` or `!$acc serial` will update the variable with some garbage values instead.

johnmauff: I think I understand the need to change the other loops.  I don't understand the need to change this loop.  The values are never used by the CPU.  Have you confirmed that changes to this particular loop are necessary for correctness, or are you just changing all instances of acc serial?

sjsprecious: Aha, I see. I just changed all the instances of `acc serial` without a `do` loop associated. I could revert this one to the original `acc serial` and double check if it matters.

sjsprecious: Ok, so reverting this line back to `acc serial` seems ok (though I do not understand why it works since there is no `do` loop here). I am trying `acc kernel` now and see if I could avoid the use of `acc update` for the other places.

sjsprecious:
After fixing the GPU bug related to `acc serial`, the verification result is closer to the previous PR https://github.com/george-bryan/CM1/pull/131 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with `testcase=15`:

# Metrics
36 stat variables are identical.
10 stat variables show a mean relative difference <= 1e-06

# 0th-order diagnostics

# 1st-order diagnostics

src/comm_droplet.F
sjsprecious: @johnmauff Actually `acc kernels` also works and I think it makes more sense than using `acc update`. I just pushed those changes.

sjsprecious (approved):

src/comm_droplet.F
johnmauff: Thanks for updating the code with the acc kernels

johnmauff (approved):
Thanks @drichte2 and @sjsprecious for working through this PR.

PR #134, opened 2023-10-08
Update s10, th10, q10

drichte2:
Moved some things around, and for testcase=15 use the interpolated 10-meter values rather than MO


sjsprecious:
The verification result is better than the previous PR #133 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with `testcase=15`:

# Metrics

# 0th-order diagnostics

# 1st-order diagnostics

sjsprecious (approved):
Thanks @drichte2 for working on this PR. It looks good to me.

I just added an OpenACC directive to update `zh` variable on the CPU end before it is used for the index calculation. This is not necessary for ASD runs since the values of `zh` do not change. But in case you are working on a case in the future where `zh` is changed on the GPU, this change will become necessary.

johnmauff (approved):
Thanks for the pull request.  This looks great.

PR #135, opened 2023-10-14
Droplet updates

drichte2:
Began as an addition of some extra droplet statistics. Then turned into a more extensive edit of the droplet_injection routine based on observed cpu/gpu differences.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -156,7 +156,7 @@ endif
 ifeq (${FC},ifort)
     # replace -xHost with -march=core-avx2 to make sure the same code builds on
     # both Intel and AMD CPUs 
-    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
     CPP          += cpp -C -P -traditional -Wno-invalid-pp-token -ffreestanding
     ifeq (${USE_OPENMP_L},true)
         OPTS     += -qopenmp
--- a/src/Makefile
+++ b/src/Makefile
@@ -156,7 +156,7 @@ endif
 ifeq (${FC},ifort)
     # replace -xHost with -march=core-avx2 to make sure the same code builds on
     # both Intel and AMD CPUs 
-    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
     CPP          += cpp -C -P -traditional -Wno-invalid-pp-token -ffreestanding
     ifeq (${USE_OPENMP_L},true)
         OPTS     += -qopenmp
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -3285,6 +3285,7 @@
 #endif
 
       if (myid.eq.0) print *, 'DHR3.4:',my_reintro,my_reintro+nparcelsLocalActive,num_injected,totdrops
+      print *, 'DHR3.5:',myid,my_reintro
       
 
       if (tot_after_reintro .lt. nparcelsMax) then   !Only add particles if the total number injected
--- a/src/droplet.F
+++ b/src/droplet.F
@@ -3320,6 +3320,7 @@
                call new_RandomNumberSequence(randomNumbers, local_seed)
                rand = getRandomReal(randomNumbers)
                a = totdrops*rand
+               a = dble(j)/dble(my_reintro)-1.0d-8
 
                !Set the droplet radius based on the sea spray generation function (SSGF)
                !Loop over all of the discretized bins of the SSGF. Based on the random number, choose which bin to draw the radius from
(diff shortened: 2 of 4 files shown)


src/droplet.F
```diff
@@ -876,13 +872,13 @@ subroutine droplet_driver (dt,dbldt,mtime,xh,uh,ruh,xf,yh,vh, &
                pdata(np,prv) = vval
                pdata(np,prw) = wval
                pdata(np,prt) = tval
-               pdata(np,prtp) = tval
+               !pdata(np,prtp) = tval
```
sjsprecious: remove this line.

src/droplet.F
```diff
@@ -3146,62 +3176,63 @@ subroutine droplet_injection(pdata,xf,yf,time_since_inject,mtime,my_reintro)
...
+
+      double precision :: rmin10_ssgf,rmax10_ssgf
       integer :: local_seed
 
-      rmin10_ssgf = log10(2e-06)
-      rmax10_ssgf = log10(500e-06)
-      dh_ssgf = 0.0242
+      rmin10_ssgf = log10(2d-06)
+      rmax10_ssgf = log10(500d-06)
+      !dh_ssgf = 0.0242
```
sjsprecious: remove this line.

src/droplet.F
```diff
@@ -3216,20 +3247,33 @@ subroutine droplet_injection(pdata,xf,yf,time_since_inject,mtime,my_reintro)
...
+      yrange = dy*ny
+
+      my_reintro_tmp = xrange*yrange*time_since_inject*totdrops/drop_mult
+#ifdef MPI
+      my_reintro = floor(my_reintro_tmp/sngl(numprocs))
+
+      !Processors 0 through the remainder-1 will create an extra particle
+      if (myid .lt. floor(mod(my_reintro_tmp,sngl(numprocs)))) then 
+         my_reintro = my_reintro + 1
+      end if
```
sjsprecious, line 3256: what is the purpose of doing this?

sjsprecious (changes requested):
Thanks @drichte2 for working on this PR. The verification result is slightly worse than the previous PR #134 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with `testcase=15`:

# Metrics

# 0th-order diagnostics

# 1st-order diagnostics

src/droplet.F
drichte2: Removed

src/droplet.F
drichte2: Removed

src/droplet.F
drichte2: The number of new droplets to inject is being computed in `my_reintro_tmp`. An issue that made the actual number of new droplets in the code depend on number of MPI ranks is due to the neglected mod(my_reintro_tmp,numprocs). Rather than put all of these remaining particles onto a single rank, I'm having ranks 1:mod all inject one extra.

sjsprecious: Thanks for the clarification.

sjsprecious:
With the new commits, the verification result is still slightly worse than the previous PR https://github.com/george-bryan/CM1/pull/134 (Derecho; CPU: ifort, 2 MPI ranks; GPU: nvhpc, 2 MPI ranks + 2 A100 GPUs), using the `namelist_ASD.verify` with `testcase=15`:

# Metrics

# 0th-order diagnostics

# 1st-order diagnostics

src/Makefile
```diff
@@ -156,7 +156,7 @@ endif
 ifeq (${FC},ifort)
     # replace -xHost with -march=core-avx2 to make sure the same code builds on
     # both Intel and AMD CPUs 
-    OPTS         += -g -O2 -march=core-avx2 -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
+    OPTS         += -g -O2 -xHost -ip -assume byterecl -fp-model precise -ftz -init=snan,arrays -traceback -no-fma
```
sjsprecious: do not use `-xHost` as it does not work on Derecho.

sjsprecious (changes requested):

sjsprecious (approved):

src/droplet.F
```diff
@@ -1076,10 +1071,21 @@ subroutine BE_integration(dt,dbldt,np,pdata,pdata_locind,xf,yf, &
             pdata(np,pract) = -1.0  !Signal this as no longer alive
             num_fallout = num_fallout + 1
             !Collect some statistics on those falling out:
+
             !$acc atomic update
             num_destroyed = num_destroyed + pdata(np,prmult)
+
             !$acc atomic update
```
johnmauff: I don't understand why we are using an atomic update here.  Isn't this an acc reduction operation?

johnmauff (commented):
While I know that acc atomic operations were not added in this commit, I am curious why they are being used instead of using an reduction directive

src/droplet.F
sjsprecious: Yes, this is a reduction operation. My concern is that this operation is done inside `BE_integration` routine, while the loop of `np` is done inside the `droplet_driver` routine, which calls the `BE_integration` routine per loop iteration. I am not sure if the acc reduction clause works properly in this case.

sjsprecious: @johnmauff I have run a new GPU test on Derecho with the `acc reduction` clause rather than the `acc atomic` directive. The simulation finishes successfully and the verification result remains the same. However, the affected variables (e.g., `num_destroyed`) are not included in the droplet diagnostic. Maybe @drichte2 can help look at the new code and see if the results are ok?

johnmauff (approved):
Thanks @sjsprecious  for looking at the atomic versus reduction issue.

